PR #6, opened 2018-04-18
initial check in for a gang of cluster creation and general configura…

MBishop17:
…tion docs.


content/rancher/tasks/global-configuration/authentication/configure-user-authentication.md
```diff
@@ -0,0 +1,87 @@
+# Configuring User Authentication
+
+You have three options for user authentication in {{< product >}}:
+
+-	**Active Directory**: Enterprises can use Active Directory (AD) for authentication, allowing users to sign in using their corporate credentials.
+
+-	**GitHub**: Open source projects or organizations that use GitHub for source control may prefer that users sign in using their GitHub accounts.
+
+-	**Basic Authentication**: If you don't want to use external authentication, you can always add users directly to {{< product >}}. We recommend using external authentication over basic authentication.
```
deniseschannon: Not sure if we we recommend external, but we should make a note that you can mix basic authentication with any of the external ones.

content/rancher/concepts/authentication.md
```diff
@@ -0,0 +1,12 @@
+# Authentication
+
+One of the key features that {{< product >}} adds to Kubernetes is enhanced user authentication.
```
will-chan: Let's call this out as centralized user auth where we can use the same auth for all k8s clusters managed by Rancher.

content/rancher/concepts/authentication.md
```diff
@@ -0,0 +1,12 @@
...
+
+By default, Kubernetes has some authentication shortcomings. Namely, users and groups are tied to the two scopes that Kubernetes uses to group resources: global and namespace. These scopes make it difficult to manage users and groups across different clusters and namespaces. Kubenetes allows integration with other authentication providers that work around these issues using plugins. However, this integration is not available to those leveraging cloud-based Kubenetes offerings such as Google Container Engine (GKE).
```
will-chan: Let's not talk about k8s shortcoming.s.  Let's say that we are enhancing/adding to it.  Also, you are mixing authention with authorization here.

content/rancher/concepts/authentication.md
```diff
@@ -0,0 +1,12 @@
...
+
+Rancher improves on Kubernetes authentication by integrating external authentication. Out-of-the-box, Rancher provides integration with the following authentication services:
```
will-chan: I think we should concentrate on pointing out that we provide centralized auth.  We are not really improving it because k8s already has a way to integrate external auth.  In fact, they have more integration than we do at this point.  We don't leverage that existing framework because it would force us to reconfigure that auth engine each time per k8s cluster.  Instead we do it centrally through Rancher with Rancher as the proxy auth engine for all k8s requests.

content/rancher/concepts/users-and-roles.md
```diff
@@ -0,0 +1,43 @@
+# Users and Roles
+
+Every user in {{< product >}} is assigned a _role_, which is a set of permissions that determine what Kubenetes API endpoints that the user can access. Let's use this metaphor: a user account is like getting access to an entire office building; however, the user's role determine what rooms the user can access within the building.
```
will-chan: misspelled kubernetes.

content/rancher/concepts/users-and-roles.md
```diff
@@ -0,0 +1,43 @@
...
+
+Within {{< product >}} there are two types of user roles:
+
+-	[Default roles](#default-roles)
+-	[Custom roles](#custom-roles)
+
+In addition, [_Membership_](#membership) plays a part in user access. Membership determines the scope of a user's role.
```
will-chan: I think i know what you mean here but doubt users will.

content/rancher/tasks/global-configuration/authentication/configure-user-authentication.md
vincent99: "Basic" Auth has a specific meaning in HTTP and is not this.  It's called "Local" in the UI.

content/rancher/installation/requirements.md
```diff
@@ -0,0 +1,50 @@
+# Rancher Requirements
+
+#### Hardware Requirements
+
+-	Memory: 4GB
+
+#### Software requirements
+
+-	Operating System: Ubuntu 16.04 (64-bit)
```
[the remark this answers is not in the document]
MBishop17: Couple of questions here:

- Do we only support 64-bit architecture?
- Does RancherOS have different versions?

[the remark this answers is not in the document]
deniseschannon: Yes, only amd64 and currently, we're only going to be supporting RancherOS 1.3.0

content/rancher/installation/requirements.md
```diff
@@ -0,0 +1,50 @@
...
+-	Software: Docker
+
+	<a name="node-requirements"></a>**Supported Versions:**
```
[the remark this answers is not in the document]
MBishop17: So here's what I did. I broke the requirements into three server different files: 

- node (node-requirements.md)
- container-based install (server-requirements.md)
- existing cluster-based install (server-requirements-deployment-install)

We're probably going to need to zoom after the review (pretty sure I still have a lot wrong), but at least this way, users won't have to think too hard about their particular requirements.

PR #224, opened 2018-06-05
Added secrets and registry sections

soumyalj:
@kinarashah : Please review

### DIFF
--- a/content/rancher/v2.x/en/tasks/projects/_index.md
+++ b/content/rancher/v2.x/en/tasks/projects/_index.md
@@ -34,12 +34,77 @@ Coming Soon
 Coming Soon
 
 ## Using Private Registries
+Private registries allow us to have repositories to hold containers that need to be private.
+Private registries can be added in the UI and images in the registry can be used
+while creating workloads. Registry can be added by providing a name, specifying
+the scope and credentials for the chosen repositry. The scope of the registry can 
+be the one which is available to all namespaces or which is available to a 
+single namespace. 
 
-Coming Soon
+In the UI,   
+1. Click on 'Add registry'  
+2. Provide a name  
+3. Choose the scope of the registry    
+4. Choose the Address [DockerHub, Quay.io, Custom]   
+5. Provide credentials of the registry chosen   
+6. Click 'Save'  
+
+Once a registry is added in rancher, the images in the registry can be chosen to
+create a workload
+
+a. From the workloads section, click on deploy a workload  
+b. Provide an image which is in the registry and choose the namespace    
+   Ex: quay.io/username/testubuntu   
+c. Click on 'Launch'. The workload should get deployed successfully
+
 
 ## Adding a Secret
 
-Coming Soon
+Secrets are used to hold sensitive information such as passwords.
+Secret can be created in rancher by providing a name along with a keyvalue pair. 
+
+The scope of the secret can be the one which is available to all namespaces or
+which is available to a single namespace. 
+A secret which is created as available to all namespaces can be used to create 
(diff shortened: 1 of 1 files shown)


content/rancher/v2.x/en/tasks/projects/_index.md
```diff
@@ -34,12 +34,77 @@ Coming Soon
...
+In the mount point provided while creating the workload, the secret keys can be
+seen as files. The contents of the file will be the value of the secret
+
+2. A workload can be created by using secret as an env variable
+
+a. From the workloads section, click on 'Deploy'   
+b. Provide a docker image and choose the namespace   
+c. Click on 'Environment variables' and 'add from source'
+d. Choose the type secret, source and keys. All keys or specific keys can be chosen   
+e. Prefix or alias can also be provided
```
kinarashah: we should mention that if not provided, default prefix for `all` keys would be `""` and default for a given key would be the same as that key name

content/rancher/v2.x/en/tasks/projects/_index.md
```diff
@@ -34,12 +34,77 @@ Coming Soon
...
+a. From the workloads section, click on 'Deploy'    
+b. Provide a docker image and choose the namespace  
+c. Click on volumes   
+d. Provide a volume name and default mode   
+e. Choose a secret which has been added   
+f. Add a mount point (Eg: /test) and subpath if required   
+g. Launch the workload   
+
+
+Once the workload is deployed, right click on the workload and execute the shell.
```
kinarashah: not sure if we need to mention clicking `vertical ellipsis` next to workload

content/rancher/v2.x/en/tasks/projects/_index.md
```diff
@@ -34,12 +34,77 @@ Coming Soon
...
+
+2. A workload can be created by using secret as an env variable
+
+a. From the workloads section, click on 'Deploy'   
+b. Provide a docker image and choose the namespace   
+c. Click on 'Environment variables' and 'add from source'
+d. Choose the type secret, source and keys. All keys or specific keys can be chosen   
+e. Prefix or alias can also be provided   
+g. Launch the workload   
+
```
kinarashah: we should also mention how to see those values, similar to how we did in volume example above

content/rancher/v2.x/en/tasks/projects/_index.md
```diff
@@ -34,12 +34,77 @@ Coming Soon
 Coming Soon
 
 ## Using Private Registries
+Private registries allow us to have repositories to hold containers that need to be private.
```
kinarashah: hold `images` that need to be private maybe?

content/rancher/v2.x/en/tasks/projects/_index.md
```diff
@@ -34,12 +34,77 @@ Coming Soon
...
+4. Choose the Address [DockerHub, Quay.io, Custom]   
+5. Provide credentials of the registry chosen   
+6. Click 'Save'  
+
+Once a registry is added in rancher, the images in the registry can be chosen to
+create a workload
+
+a. From the workloads section, click on deploy a workload  
+b. Provide an image which is in the registry and choose the namespace    
+   Ex: quay.io/username/testubuntu
```
kinarashah: since this is a known issue, we should document that in case of some custom address, it should be the same as the fqdn of your private registry.  (should be exact full path given in step 4 above while creating registry) 

also should mention this will only work if workload is created in UI. not through `kubectl`. I wrote it in `concepts` section but reminder should be better since we already had users submitting issue about this

PR #227, opened 2018-06-05
AD search base and search attribute

mrajashree:
Configuring AD and how to use search base for account name, first name, last name


deniseschannon:
Needs some formatting love.

@MBishop17 I think we should separate out each auth provider into it's own page so it can have it's own search when users are looking.

PR #223, opened 2018-06-05
Docs for task: Creating Projects

prachidamle:
@soumyalj please review the steps for Creating a Project

PR #229, opened 2018-06-05
Docs for task: Switching Projects

prachidamle:
@soumyalj please review


deniseschannon:
@MBishop17 we should make this into a subsection of projects with https://github.com/rancher/docs/pull/223

but i'm wondering if we should even really document this....

MBishop17:
@deniseschannon, I'm going to go ahead and document this. I don't think the control for switching between clusters/projects is intuitive from a UX perspective. I can see some people needing help with this.

PR #216, opened 2018-06-05
Cluster memberships

mrajashree:
Tasks: Managing cluster membership
Tasks: Adding users to clusters, talk about how groups works

PR #231, opened 2018-06-06
Docs for Tasks: Adding/Managing  members to projects

prachidamle:
@soumyalj review please.


soumyalj:
LGTM

deniseschannon:
@MBishop17 We should make this matches with whatever we format for as both concepts are almost the same. the only difference is that you're at the projectlevel and the project level has a built-in readonly role. https://github.com/rancher/docs/pull/216

in both PRs, we should make notes about how if they don't want to see the built-in roles, they should lock them and link to the locking roles PRs.

PR #287, opened 2018-06-12
Add required options for cloud providers

galal-hussein:


deniseschannon:
Let's just put an asterisk in the column for the required fields and leave it blank if it's not required.

PR #293, opened 2018-06-13
procedure to change projects

MBishop17:
related to #229


deniseschannon:
I think we could add in content that if you are in the global view, you can click on "clusters" to get back to the cluster page. and from the cluster page, you can select a cluster. 

similar if you are already in a cluster, you could go to projects/namespaces, to see all namespaces grouped in projects. and then select the project you want to navigate to.

PR #296, opened 2018-06-14
Instructions on how to add members to projects

MBishop17:
New PR for #231

### DIFF
--- a/content/rancher/v2.x/en/tasks/projects/add-project-members/_index.md
+++ b/content/rancher/v2.x/en/tasks/projects/add-project-members/_index.md
@@ -3,4 +3,47 @@ title: Adding Project Members
 weight: 
 draft: true
 ---
-Coming Soon
+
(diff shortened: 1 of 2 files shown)


content/rancher/v2.x/en/tasks/projects/add-project-members/_index.md
```diff
@@ -3,4 +3,47 @@ title: Adding Project Members
 weight: 
 draft: true
 ---
-Coming Soon
+
+Before a user can access a project and its resources, the project owner must explicitly assign that user a project _membership_, which provides the user with access and roles within the project.
```
deniseschannon: We should differentiate between cluster membership and project membership. Explain why you would want to give cluster membership versus project membership.

cluster membership - access to all projects in a cluster and nodes in the cluster
project membership - access only to the namespaces in the projects, no access to nodes or other projects

PR #372, opened 2018-06-29
Update default network provider.

drewgwallace:
Default when using `./rke config` and also described here:
https://rancher.com/docs/rke/v0.1.x/en/config-options/add-ons/network-plugins/


deniseschannon:
@moelsayed @galal-hussein In RKE CLI, do we default to flannel or canal?

@drewgwallace In Rancher, we default to canal through the UI but not sure if that's the case in RKE CLI.

galal-hussein:
@deniseschannon in RKE cli we default to canal: https://github.com/rancher/rke/blob/master/cluster/defaults.go#L29

PR #377, opened 2018-06-30
Update _Index.md

gabx:
1. Code inside a list must be inside <code>  code </code>
2. The command line `cat MyCert | base64 -d` let you quickly verify the certificate integrity.


deniseschannon:
@gabx Typically, when you put items in single mark ticks which make code blocks. As hugo reads the markdown files and converts it. 

Can you update to use them instead of the HTML? 

```
`tls.crt: LS01...`
```

gabx:
not sure to understand what you are asking me

PR #379, opened 2018-07-01
update configure Load Balancer

gabx:
The provided `nginx.conf` will not work as it in all distro.
1- you must tell nginx at the top of the config file to load the stream module
2- you must have the nginx module package installed.
For example, on fedora 28, this is a separate package from nginx.

### DIFF
--- a/content/rancher/v2.x/en/installation/ha-server-install/_index.md
+++ b/content/rancher/v2.x/en/installation/ha-server-install/_index.md
@@ -119,9 +119,9 @@ After installing NGINX, you need to update the NGINX config file, `nginx.conf`, 
     worker_processes 4;
     worker_rlimit_nofile 40000;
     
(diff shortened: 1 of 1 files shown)


superseb:
Thanks for the PR, can you change:

- Make the `load_module` commented out by default, as it's usually an exception and paths are too specific which will break other distributions
- Put the note also in the comments so people know what it is related to
- Make the note specific to the distro(s) you tested on, and fix the `directoty` typo
- Package name for Fedora seems to `nginx-all-modules`

superseb:
Thanks for the contribution.

We decided to not include every OS's options in the docs, I've created a new PR here which describes the requirements for NGINX: https://github.com/rancher/docs/pull/396.

MBishop17:
Closing. Resolved with #396 .

PR #387, opened 2018-07-03
Add SSL FAQ on PKCS8

superseb:


vincent99:
I know this addresses a problem, but this is a ridiculous problem to be putting into the FAQ instead of adding 5 lines to support pkcs8 directly...

superseb:
It's not supported now, so it needs to be addressed. I wanted to fix it in the code later.

Issue #75, opened 2018-05-03
Describe cleaning/deleting node


superseb:
Need to describe cleaning a node when re-used or deleted.

moelsayed:
RKE tries to minimize the changes on host nodes, so to clean up, we run a simple list of commands:
```
# clean up docker containers, volumes and images
docker rm -f $(docker ps -aq)
docker volume rm -f $(docker volume ls -q)
docker rmi -f $(docker images -q)

# remove files/directories we write to disk
rm -rf /var/lib/etcd/
rm -rf /etc/kubernetes/
rm -rf /etc/cni/
rm -rf /opt/cni/
rm -rf /var/lib/cni/
rm -rf /var/run/calico/

# reboot!
reboot
```

PR #430, opened 2018-07-10
Add cleaning cluster nodes

superseb:
https://github.com/rancher/docs/issues/75

### DIFF
--- a/content/rancher/v2.x/en/installation/removing-rancher/cleaning-cluster-nodes/_index.md
+++ b/content/rancher/v2.x/en/installation/removing-rancher/cleaning-cluster-nodes/_index.md
@@ -1,0 +1,186 @@
+---
+title: Cleaning cluster nodes
+weight: 375
+---
+When adding a node to a cluster, resources (containers/(virtual) network interfaces) and configuration items (certificates/configuration files) are created. When removing a node from a cluster (if it is in `Active` state), those resources will be automatically cleaned and the only action needed is to restart the node. When a node has become unreachable and the automatic cleanup process cannot be used, we describe the steps that need to be executed before the node can be added to a cluster again.
+
+## Removing a node from a cluster via Rancher UI
+
+When the node is in `Active` state, removing the node from a cluster will trigger a process to clean up the node. Please restart the node after the automatic cleanup process is done to make sure any non-persistent data is properly removed.
+
+* How to restart a node
+
+```
+# using reboot
+reboot
+
+# using shutdown
+shutdown -r now
+```
+
+## Cleaning a node manually
+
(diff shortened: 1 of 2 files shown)


content/rancher/v2.x/en/installation/removing-rancher/cleaning-cluster-nodes/_index.md
```diff
@@ -0,0 +1,145 @@
...
+`/var/log/containers` |
+`/var/log/pods` |
+`/var/run/calico` |
+
+* How to clean the directories:
+
+```
+rm -rf /etc/ceph \
+       /etc/cni \
+       /etc/kubernetes \
```
galal-hussein: /etc/kubernetes
/var/lib/etcd/
/etc/cni/
/opt/cni/
/var/lib/cni/
/var/run/calico/
/etc/kubernetes/.tmp/

all of these previous directories are being cleaned by `rke remove` there is no need to include them

content/rancher/v2.x/en/installation/removing-rancher/cleaning-cluster-nodes/_index.md
```diff
@@ -0,0 +1,145 @@
+---
+title: Cleaning cluster nodes
+weight: 375
+---
+
+When adding a node to a cluster, the following items are created/configured on the node:
```
galal-hussein: i think its important to mention that when **removing a cluster** rke will clean each host if it can connect to it mentioning the what rke does on cleaning including cleaning directories and k8s components

content/rancher/v2.x/en/installation/removing-rancher/cleaning-cluster-nodes/_index.md
```diff
@@ -0,0 +1,145 @@
...
+ip link delete interface_name
+```
+
+<h3>Iptables</h3>
+
+Iptables rules are used to route traffic from and to containers. The created rules are not persistent, so restarting the node will restore iptables to it's original state.
+
+>**Warning:** The commands listed below may break default node behavior. The recommended way is to restart the node to restore the original state of the iptables rules.
+
+* How to remove all iptables rules
```
galal-hussein: these are really dangerous set of commands and it can block the user from the machine if applied in a wrong way, and will break docker as far as i can see, because you are removing all the nat rules including docker's nat rules
so its best to just restart the machine

galal-hussein: i don't think we should mention these rules even with a warning

superseb:
- Added two scenarios, removing node when active (automatic cleaning) or manual. Details need to be included for manual, as you can't rely on the fact that the node is always reachable.
- Removed iptables commands to remove/flush

superseb:
@deniseschannon ^^

Issue #325, opened 2018-06-20
Configuring Active Directory Authentication - insufficient documentation


CrowdfoxMKL:
Hello,

my colleague and me are trying to get the Active Directory Authentication up and running with our Univention UCS as an AD/LDAP provider.

Unfortunately we can't get it to work, but that's not the matter here. 
One problem we are facing is the insufficient documentation to be found [here](https://rancher.com/docs/rancher/v2.x/en/tasks/global-configuration/authentication/active-directory/)

No mentions in which format the username is to be entered (in DN-notation or as a simple username) and generally no trouble shooting help there. 

It would be great if someone tackled this problem, to help people new to rancher or external authentication in general save some frustrating hours trying to get this working with the information given.

MBishop17:
Hi @CrowdfoxMKL,

I'm the technical writer who owns the Rancher docs repo. I'm sorry that our docs don't have the information that you need to setup AD! I know how frustrating that lack of information can be. I'll work with our developers to get this area of the docs improved. Again, my apologies, and thanks for filing this issue so that we can address it.

Mark Bishop
Technical Writer @ Rancher

CrowdfoxMKL:
Hi @MBishop17,

thanks for your reply, I think I might be able to help a bit further by clarifying some points. After we got it to work with an OpenLDAP based AD replacement, we found out about some subtle differences between the original Microsoft implementation and OpenLDAP.

1.) Microsoft AD has a backward compatibility mode enabled, which still accepts the format DOMAIN\User. OpenLDAP (at least in the Univention UCS and the one from Ubuntu) seems not to be able to parse this format by default.
So, to use an OpenLDAP server, one has to leave the field "Default Logon Domain" has to be empty - or else Rancher will try to use the "old" format.

2.) If you want to be able to use users with the same username from different domains resting on the same LDAP server, you still have to give the domain name somehow. This can be achieved quite easily.
In the field "Service Account Name" enter the user in the form User@domain (eg. user@example.domain) and under "Customize Schema" enter "userPrincipalName" into the field "Search Attribute" under "Users" - this field in LDAP/AD holds the complete user with his domain in exactly that format.

3.) It seems, that some subtle differences are also present in the LDAP directory itself. To be able to login, we had to modify the field "User Search Base" in the first block, this has to read "cn=users,dn=example,dn=domain" or similar. The point here is to use "cn=users" as opposed to "ou=users" which is needed for an MS AD.

[...]

PR #489, opened 2018-07-19
Revamped AD auth documentation

janeczku:
See #325


leodotcloud:
Thanks for adding this info! Looks good.

PR #519, opened 2018-07-24
Small link fix

MBishop17:
fixed the two remaining links.

Not sure what the deal is with the third unchecked link is (the one marked ***duplicate?***). It's the same link as the one above it.


deniseschannon:
LGTM

Issue #267, opened 2018-06-08
More ports need to be mentioned


leodotcloud:
When provisioning a machine from the UI, there are more ports added to the security group by docker-machine. See [here](https://docs.docker.com/machine/drivers/aws/#security-group).

Need to update the picture here: https://rancher.com/docs/rancher/v2.x/en/installation/ha-server-install/

2376 is added by docker-machine.

PR #530, opened 2018-07-25
Add info on rancher-nodes security group when using EC2 node driver

superseb:


deniseschannon:
LGTM

deniseschannon:
https://github.com/rancher/docs/issues/267

Issue #537, opened 2018-07-26
Add proxy info for HA


superseb:
Current proxy info is for single, needs test/update for HA

PR #540, opened 2018-07-26
Add configuring HTTP proxy for HA

superseb:
https://github.com/rancher/docs/issues/537


superseb:
If you set `no_proxy` only, network ranges set in them won't work. So I left them out entirely as it will only confuse users.

jgreat:
I'm good with that.  The content LGTM

PR #573, opened 2018-08-02
2.0.7 Documentation: Default Global Permissions and Roles

MBishop17:


dramich:
LGTM

deniseschannon:
@MBishop17 can we add in `_Available as of v2.0.7_` for where we talk about how to configure it out. Or some language how you can configure it as of v2.0.7?

MBishop17:
added _available as of_ text

PR #585, opened 2018-08-05
Fix chapter description

stieler-it:


deniseschannon:
Merged the same fix in #612. Thanks @stieler-it!!

PR #606, opened 2018-08-10
Fixes a mistake on the installation overview page

adjika:
`Backups and Rollbacks` mistakenly had the same description as `Port Requirements`:
`This page lists the ports you must open to operate Rancher.`


deniseschannon:
Thanks @adjika , I just merged in #612 that fixes this!

PR #612, opened 2018-08-12
Fix wording of 'backups and rollbacks' subtitle

ReidWeb:
The existing subtitle does not correlate with the section, looks like it was copied from elsewhere and not updated


deniseschannon:
Thanks @ReidWeb

PR #621, opened 2018-08-15
changing the order of checking after upgrade

urgemerge:
changing the documented order of checking if an upgrade has succeeded


deniseschannon:
Thanks @urgemerge!

PR #638, opened 2018-08-21
add nginx proxy parameter

brechsteiner:


superseb:
Thanks for your contribution. Can you describe what issue is solved here (how to reproduce and what you run into)?

brechsteiner:
it was not possible to use kubectl logs --follow=true .... the command just hangs without any output

superseb:
It does not reproduce if it's the same NGINX config posted in the other issue, what kubectl did you use and do you see it with the embedded kubectl as well?

brechsteiner:
kubectl version: v1.11.2
The web-embedded kubectl works as expected, the rancher-cli-embedded kubectl doesn't work for me, i get this error:
```
rancher kubectl get nodes
Unable to connect to the server: x509: certificate signed by unknown authority
FATA[0001] exit status 1
```

deniseschannon:
@brechsteiner We just moved what ws in our `staging` branch into `master` as we introduced a new helm install for HA, which moved some files around.

PR #643, opened 2018-08-22
Acme challenge verify not working

insekticid:
Fix for self signed certs - validation works only over http

### DIFF
--- a/content/rancher/v2.x/en/installation/ha-server-install/_index.md
+++ b/content/rancher/v2.x/en/installation/ha-server-install/_index.md
@@ -120,9 +120,24 @@ After installing NGINX, you need to update the NGINX config file, `nginx.conf`, 
     }
 
     http {
+        upstream rancher_servers {
+            least_conn;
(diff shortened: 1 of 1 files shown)


superseb:
Thanks for your contribution. Lets Encrypt is not documented and not supported in HA setup with the built in support (it will work but will have issues when it is scaled). Also, the documentation is on using NGINX as L4 using stream.

insekticid:
I have deployed cert manager and it works well with this setup. You will not have this issue when you will use DNS01 Challenge

superseb:
Like mentioned above, it will work, but neither adding cert-manager or using Lets Encrypt certificates is documented in this part. There will be new docs soon using Helm chart install and the possibility to use cert-manager.

deniseschannon:
@insekticid We just moved what was in our `staging` branch into `master` as we introduced a new helm install for HA, which moved some files around. 

Sorry, you'll need to rebase as the files you referenced have moved.

deniseschannon:
This is no longer relevant as the helm chart install has been made the standard for HA installs.

Issue #664, opened 2018-08-24
revisit single node install/upgrade/rollback/backup/restore


deniseschannon:
now that the commands are correct, we need to add in more context so people can understand why and what they are doing and how the docs flow between each other.

deniseschannon:
**install for single node and single node with external LB**
* fix the software requirements section (too many notes)
* make the requirements section consistent (move to tabs style)

**single node install/single node install with external LB/upgrade**
* fix the air gap/audit log notes into an advanced options section (maybe shortcode to be used in the different install options

**single node install**

certs issue in option b:
<img width="623" alt="screen shot 2018-08-28 at 11 25 27 am" src="https://user-images.githubusercontent.com/11410997/44742860-cf636400-aab5-11e8-9c5e-eecb91b49e5b.png">

certs issue in option c:

<img width="517" alt="screen shot 2018-08-28 at 11 26 53 am" src="https://user-images.githubusercontent.com/11410997/44742869-d38f8180-aab5-11e8-94b6-8db17b18b63c.png">

In the what's next section, it's weird to have them create a backup cause they just installed. we should recommend that they review how to take backups, but no reason to take a backup.

deniseschannon:
Available in live now with PR #709

PR #709, opened 2018-08-28
making single node install, upgrade/rollback, backup/restore clearer

deniseschannon:


deniseschannon:
https://github.com/rancher/docs/issues/664

PR #715, opened 2018-08-29
Add section and link to custom CA and proxy

superseb:


deniseschannon:
@superseb I don't understand why custom certs needs to be bumped to a higher level.

superseb:
It's technically wrong to have this in the list of SSL options as it's not an option to front Rancher with SSL, but to validate chains to outgoing connections from Rancher. As this was explained to @jgreat in https://github.com/rancher/rancher/pull/14945#issuecomment-417404121, maybe he can say why he moved it to this location as I added it as a separate section when I created it.

jgreat:
@superseb is correct. That section doesn't belong in the single node install doc. I think I put that doc under the single node install tree when we were moving things around because it only has instructions for doing it with a single node install. Not sure how it got merged into the install doc.

Its placement makes it a bit confusing right now.  I'd move it out of the main install tree and rename it something like "How to Install Additional CA Certificates". Make it a reference doc with instructions for both single node and ha installs.

PR #725, opened 2018-08-29
Adding blog descriptions for 1.6 to 2.0 migration

prachidamle:
@deniseschannon added the descriptions for each blogs published so far.


deniseschannon:
@MBishop17 can you review for grammar and tone? As this is on the docs site, we should be consistent with what else we've written for 2.0. 

Eventually, we'll want to convert the blogs into formal docs of how to migrate, but this is the first step.

deniseschannon:
Closing in favor #754

PR #741, opened 2018-09-03
Revamped port requirements doc

janeczku:


deniseschannon:
@jgreat can you help review for content that this make sense?

PR #745, opened 2018-09-05
making procedure for recommended way of deleting catalog apps

MBishop17:


content/rancher/v2.x/en/catalog/_index.md
```diff
@@ -95,3 +95,19 @@ After you've either enabled the built-in catalogs or added your own custom catal
 
 - **Workloads** view
 - **Catalog Apps** view
+
+## Deleting Catalog Application Deployments
+
+When you want to delete a deployed catalog application, we recommend deleting both its namespace and the catalog app itself. This practice ensures no clutter is left behind when you delete the catalog app, as merely deleting the app leaves behind an orphaned namespace.
```
deniseschannon: We should just say that we recommend deleting the namespace (this automatically deletes the catalog app). 

Also, we should mention the reason why deleting the app itself will not delete the namespace as other workloads/apps may have been deployed in the namespace. I don't think we need to call it an `orphaned namespace`.

MBishop17:
lives at https://rancher.com/docs/rancher/v2.x/en/catalog/

PR #746, opened 2018-09-05
User should be "Standard User"

steakfest:
The first couple references to the global permission "Standard User" just said "User"  

I'm just learning Rancher, but this seemed like a mistake to me.


MBishop17:
@deniseschannon, this change lgtm.

PR #748, opened 2018-09-05
removing redundant content (you were right Denise)

MBishop17:
Got feedback that the top-level content for custom nodes was confusing. Removed it since it's documented in the RKE section.


deniseschannon:
most people don't know what a RKE cluster is which is why we might want to keep a custom cluster top level section.

Issue #527, opened 2018-07-25
Project Documentation: Define Quotas


MBishop17:
- Explain what project quotas are.
- Explain how to create them.

PR #793, opened 2018-09-13
Resource Quota Draft

MBishop17:
Alena, can I please get some feedback on the draft I wrote for Resource Quotas?

You can view the actual draft page at https://staging.rancher.com/docs/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/project-quotas/


alena1108:
> resources being things like data or processing power.

can we drop "data" from the example as we are not exposing storage quota in the UI? Only cpu/memory/resource limits (like number of pods) are available


> Resource quotas in Rancher work similarly to how they do in the native version of Kubernetes. 

We should mention that Rancher implementation leverages native Kubernetes Resource Quota concept.

> In a standard Kubernetes deployment, resource quotas are applied to individual namespaces.
> Quotas are applied to individual namespaces, rather than collectively. Even though each quota sets the same limits, a unique quota is applied to each namespace.

`Even though each quota sets the same limits` - We should perhaps be more clear on the use case we are trying to solve. That sometimes admin wants the same resource quota to be applied to multiple namespaces. Also from administrator perspective, it would be nice to define the quota on the project, so every namespace created in the project would have the default resource quota set, so user doesn't have to enter it.

> When the overall limit for the project is exceeded, Kubernetes uses logic to determine which namespaces to stop to get back under the quota.

Not quite true. We enforce the limit on the API level by disallowing the namespace creation if its resource quota exceeds the limit available on the project. We do not stop the existing namespaces.

> If an individual namespace exceeds its namespace limit, Kubernetes stops anything objects in the namespace from operating.

Same here.


More to add:

We should describe that quota defined on the namespace will get propagated to kubernetes cluster, and Kubernetes native capabilities will enforce that all the resources created within the namespace, will be within the namespace quota. Like if user tries to start the workload which CPU request exceeds the namespace limit, the workload creation would fail.


We should emphasize that on namespace creation, if the quota defined on the namespace, exceeds the quota available on the project (considering all existing namespaces and their quotas), the namespace creation would fail. Admin would have to adjust the project quota by increasing the limit.

MBishop17:
@alena1108 and @deniseschannon , I've completed a draft for resource quotas. It's a long one, so enjoy!

content/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/_index.md
```diff
@@ -171,4 +189,23 @@ Cluster admins and members may occasionally need to move a namespace to another
 
 1. Choose a new project for the new namespace and then click **Move**. Alternatively, you can remove the namespace from all projects by selecting **None**.
 
-**Result:** Your namespace is moved to a different project (or is unattached from all projects). If any project resources are attached to the namespace, the namespace releases them and then attached resources from the new project. 
+**Result:** Your namespace is moved to a different project (or is unattached from all projects). If any project resources are attached to the namespace, the namespace releases them and then attached resources from the new project.
```
alena1108: we do not allow moving namespace if the destination project has the resource quota set. Thats the limitation for now. The only moves that are allowed, are:

* from project with no quota, to the project with no quota
* from project with quota, to the project with no quota. The quota on the namespace will be cleaned up as the result

content/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/_index.md
```diff
@@ -171,4 +189,23 @@ Cluster admins and members may occasionally need to move a namespace to another
...
+
+### Editing Namespace Resource Quotas 
+
+If there is a [resource quota]({{< baseurl >}}/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/resource-quotas) configured for a project, you can override the namespace default limit to provide a specific namespace with access to more (or less) project resources.
```
alena1108: > with access to more (or less) project resources
think would be good to note that the namespace quota value should be within project limit value?

content/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/editing-projects/_index.md
```diff
@@ -69,13 +69,40 @@ You can always assign a PSP to an existing project if you didn't assign one duri
...
+1. Expand **Resource Quotas** and click **Add Quota**. Alternatively, you can edit existing quotas.
+    
+1. Select a [Resource Type]({{< baseurl >}}/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/resource-quotas/#resource-quota-types).
+     
+1. Enter values for the **Project Limit** and the **Namespace Default Limit**.
+        
+    | Field                   | Description                                                                                              |
+    | ----------------------- | -------------------------------------------------------------------------------------------------------- |
+    | Project Limit           | The overall resource limit for the project.                                                              |
+    | Namespace Default Limit | The default resource limit available for each namespace. This limit is propagated to each namespace in the project. |
```
alena1108: should mention that combined limit of all the namespaces shouldn't exceed the quota set on a project

MBishop17:
related to #527

content/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/_index.md
```diff
@@ -106,6 +106,22 @@ Rancher extends Kubernetes to allow the application of [Pod Security Policies](h
         >**Tip:** Choose Custom to create a custom role on the fly: [Custom Project Roles]({{< baseurl >}}/rancher/v2.x/en/admin-settings/rbac/cluster-project-roles/#custom-project-roles).
 
     1. To add more members, repeat substeps a—c.
+    
+1. **Optional:** Add **Resource Quotas**, which limit the resources that a project (and its namespaces) can consume. For more information, see [Resource Quotas]({{< baseurl >}}/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/resource-quotas).
```
deniseschannon: we should indicate which version this is available in.

content/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/_index.md
```diff
@@ -167,8 +185,32 @@ Cluster admins and members may occasionally need to move a namespace to another
 
 1. Select the namespace(s) that you want to move to a different project. Then click **Move**. You can move multiple namespaces at one.
 
-    >**Note:** Don't move the namespaces in the `System` project. Moving these namespaces can adversely affect cluster networking.
+    >**Notes:** 
+    >
+    >- Don't move the namespaces in the `System` project. Moving these namespaces can adversely affect cluster networking.
+    >- You cannot move a namespace into a project that already has a [resource quota]({{< baseurl >}}/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/resource-quotas/) configured.
```
deniseschannon: we should indicate that if you move a namespace that had a quota set, the quotas would be removed when moving to a project with no quota.

content/rancher/v2.x/en/k8s-in-rancher/projects-and-namespaces/resource-quotas/_index.md
```diff
@@ -0,0 +1,89 @@
...
+| Persistent Volume Claims | The maximum number of persistent volume claims that can exist in the project/namespace.                                                                                                           |
+| Replications Controllers | The maximum number of replication controllers that can exist in the project/namespace.                                                                                                            |
+| Secrets                  | The maximum number of secrets that can exist in the project/namespace.                                                                                                                            |
+
+>**<sup>1</sup>** In the quota, if you set CPU or Memory limits, all containers you create in the project / namespace must explicitly satisfy the quota. See the [Kubernetes documentation](https://kubernetes.io/docs/concepts/policy/resource-quotas/#requests-vs-limits) for more details.
+
+ 
+### Namespace Default Limit Overrides
```
deniseschannon: I'd rename this to `Overriding the default Limit for a  Namespace`

MBishop17:
@deniseschannon , are we waiting on anything to close out this issue?

deniseschannon:
I can't merge to staging as it will cause staging to be stuck in being able to be moved to live. @MBishop17

Issue #794, opened 2018-09-13
Document note around read-only project member


dramich:
If a user is added to two projects, one as read-only and one as a project-member, they will be able to create namespaces in the read-only project.

PR #795, opened 2018-09-13
read-only isn't always ready-only

dramich:
Issue https://github.com/rancher/docs/issues/794

### DIFF
--- a/content/rancher/v2.x/en/admin-settings/rbac/cluster-project-roles/_index.md
+++ b/content/rancher/v2.x/en/admin-settings/rbac/cluster-project-roles/_index.md
@@ -60,6 +60,9 @@ _Project roles_ are roles that can be used to grant users access to a project. T
 - **Read Only:**
 
     These users can view everything in the project but cannot create, update, or delete anything.
+    
+    > **Note:** If a user is added to a project as an owner or a member they will have permission to create namespaces in all projects they are a member of, even as a Read Only member. 
+    
 
 #### Custom Project Roles
 


content/rancher/v2.x/en/admin-settings/rbac/cluster-project-roles/_index.md
```diff
@@ -60,6 +60,9 @@ _Project roles_ are roles that can be used to grant users access to a project. T
 - **Read Only:**
 
     These users can view everything in the project but cannot create, update, or delete anything.
+    
+    > **Note:** If a user is added to a project as an owner or a member they will have permission to create namespaces in all projects they are a member of, even as a Read Only member.
```
cjellick: I think you should explain why. like:
> This is because creating namespaces is a cluster level permission. It cannot be controller on a per project basis.

MBishop17: @cjellick , I just discussed this with Dan, and I don't think we should get too deep into why, only that it's a Kubernetes thing:

"Because of how Kubernetes handles permissions, if a user is added to a project as an owner or a member they will have permission to create namespaces in all projects they are a member of, even as a Read Only member. "

cjellick: Fine by me. I just want some bit of "why" since it will feel like a security issue to someone that is hit by it

PR #807, opened 2018-09-17
table suggestion

lvuch:


deniseschannon:
LGTM. @MBishop17 thoughts?

MBishop17:
lgtm too.

Issue #614, opened 2018-08-13
Install Guide: Air Gap HA Install


MBishop17:
https://rancher.com/docs/rancher/v2.x/en/installation/air-gap-installation/ doesn't have instructions for HA in an air gap environment.

rawmind0:
Additionally of instructions showed in the HA docs, for air-gap installation and use private registry images, RKE template needs to be edited as: 

  - adding `private_registries` section

```
private_registries:
  - url: <YOUR_REGISTRY_URL>
    user: <USER>
    password: <PASSWORD>
```

  - Adding `system_images` section accordingly private registry url

```
system_images:
  kubernetes: <YOUR_REGISTRY_URL>/rancher/hyperkube:v1.10.3-rancher2
  etcd: <YOUR_REGISTRY_URL>/rancher/coreos-etcd:v3.1.12
  alpine: <YOUR_REGISTRY_URL>/rancher/rke-tools:v0.1.9
  nginx_proxy: <YOUR_REGISTRY_URL>/rancher/rke-tools:v0.1.9
  cert_downloader: <YOUR_REGISTRY_URL>/rancher/rke-tools:v0.1.9
  kubernetes_services_sidecar: <YOUR_REGISTRY_URL>/rancher/rke-tools:v0.1.9
  kubedns: <YOUR_REGISTRY_URL>/rancher/k8s-dns-kube-dns-amd64:1.14.8
  dnsmasq: <YOUR_REGISTRY_URL>/rancher/k8s-dns-dnsmasq-nanny-amd64:1.14.8
  kubedns_sidecar: <YOUR_REGISTRY_URL>/rancher/k8s-dns-sidecar-amd64:1.14.8
  kubedns_autoscaler: <YOUR_REGISTRY_URL>/rancher/cluster-proportional-autoscaler-amd64:1.0.0
  pod_infra_container: <YOUR_REGISTRY_URL>/rancher/pause-amd64:3.1
  flannel: <YOUR_REGISTRY_URL>/rancher/coreos-flannel:v0.9.1
  flannel_cni: <YOUR_REGISTRY_URL>/rancher/coreos-flannel-cni:v0.2.0
  calico_node: <YOUR_REGISTRY_URL>/rancher/calico-node:v3.1.1
  calico_cni: <YOUR_REGISTRY_URL>/rancher/calico-cni:v3.1.1
  calico_ctl: <YOUR_REGISTRY_URL>/rancher/calico-ctl:v2.0.0
[...]

PR #808, opened 2018-09-18
initial draft of ha air gap install

MBishop17:
Related to #614 

Initial grammar pass for @jgreat . I also have a few questions.


content/rancher/v2.x/en/installation/air-gap-installation/ha-air-gap-installation/_index.md
```diff
@@ -0,0 +1,246 @@
...
+                <th>Download</th>
+                <th>Description</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr>
+                <td>RKE</td>
+                <td>
+                    A list of RKE images that are required can be compiled by running the following command.
+                    <pre><code>rke config --system-images</code></pre>
```
MBishop17: @jgreat , I was confused by what's going on in "Gather Images".  Is the first part listing the files you need along with their purpose, and then the second part (the part with the script) downloading them?

janeczku:
We also need to document mirroring the rancher server chart repository, e.g.
1) How to create a local helm repository to host the rancher server chart from https://github.com/rancher/server-chart or 
2) Point out to add this repo to helm instead of the public one during this step https://rancher.com/docs/rancher/v2.x/en/installation/ha/helm-rancher/#add-the-chart-repo

MBishop17:
> There is already a part about mirroring images in this section, why not re-use that information? This does not make for an uniform look and feel how to do air gap for both installs.

> We also need to document mirroring the rancher server chart repository, e.g. either by
> 
> 1. Creating a local helm repository to host the rancher server chart from https://github.com/rancher/server-chart,
>    or
> 2. Probably better: How to pull the packaged archive of rancher server chart and then use helm to deploy from that local package.

@janeczku  and @superseb , can you please leave inline comments on the parts of the docs that you're calling out?

PR #862, opened 2018-10-03
Fix link to AD FS page

superseb:
Also changed `First Appearance` to `Available as of` like in the subpages.


deniseschannon:
can you make this PR from the staging branch? all the docs for 2.1 are going to be coming from there and this page has changed a bit in staging.

superseb:
Done

PR #892, opened 2018-10-12
Use name 'Projects/Namespaces' instead of 'Projects'

stefanlasiewski:
The Quickstart guide says "From the main menu of the Dashboard, select Projects.", but there is no tab named "Projects". I believe it was renamed to be 'Projects/Namespaces'.


deniseschannon:
Thanks @stefanlasiewski!

PR #943, opened 2018-10-24
Add CertManager issuer and clusterIssuer to docs

patst:
Added docs for https://github.com/rancher/rancher/issues/16178 .

The code changes are in the server helm chart ( https://github.com/rancher/server-chart/pull/40 )


patst:
Solved with the additional annotations in the helm chart, so this ist no longer needed

PR #986, opened 2018-11-13
Update _index.md

darren-rose:
an answer is required


superseb:
Thanks for the PR but there is no real point in adding errors to the FAQ without a solution. If we solve https://github.com/rancher/rke/issues/1023, the solution can be added.

PR #992, opened 2018-11-16
Fix kubernetes 1.12 version

rekcah78:
because image rancher/hyperkube:v1.12.0-rancher1-1 not found


deniseschannon:
@rekcah78 That version is not an image tag for hyperkube. That is the version that we use for a kubernetes and all it's components. 

https://github.com/rancher/rke/blob/v0.1.11/vendor/github.com/rancher/types/apis/management.cattle.io/v3/k8s_defaults.go#L530

The hyperkube image and tag is `rancher/hyperkube:v1.12.0-rancher1`

PR #1037, opened 2018-12-03
add pipeline notification configuration

gitlawr:
Address issue #1036

### DIFF
--- a/content/rancher/v2.x/en/tools/pipelines/configurations/_index.md
+++ b/content/rancher/v2.x/en/tools/pipelines/configurations/_index.md
@@ -491,7 +491,7 @@ You can configure pipelines to send notifications to [notifiers](https://rancher
 
 1. If you don't have any existing notifier, click **create one** and you will be directed to the notifiers page. Follow the [instruction](https://rancher.com/docs/rancher/v2.x/en/tools/notifiers-and-alerts/#adding-notifiers) to add a notifier. If you have already configured notifiers, click **Add Recipient** button.
 
-1. Select a notifier from the dropdown, then input recipient or use the default one. By clicking **Add Recipient** button, you can configure to send notifications to different notifiers or multiple recipients.
+1. Select a notifier from the dropdown, then input recipient or use the default one. By clicking the **Add Recipient** button, you can send notifications to different notifiers or multiple recipients.
  
 {{% /tab %}}
 {{% tab "By YAML" %}}
--- a/content/rancher/v2.x/en/tools/pipelines/configurations/_index.md
+++ b/content/rancher/v2.x/en/tools/pipelines/configurations/_index.md
@@ -477,7 +477,7 @@ branch:
 
 ### Configuring Notifications
 
-_Available as of v2.2_
+_Available as of v2.2.0_
 
 You can configure pipelines to send out notifications to any [notifiers]({{< baseurl >}}/rancher/v2.x/en/tools/notifiers-and-alerts/#notifiers) based on the build status of a pipeline. Before configuring a notification, Rancher recommends [setting up notifiers]]({{< baseurl >}}/rancher/v2.x/en/tools/notifiers-and-alerts/#adding-notifiers) so it will be easy to add recipients immediately. 
 
--- a/content/rancher/v2.x/en/tools/pipelines/configurations/_index.md
+++ b/content/rancher/v2.x/en/tools/pipelines/configurations/_index.md
@@ -477,23 +477,25 @@ branch:
 
 ### Configuring Notifications
 
+_Available as of v2.2_
+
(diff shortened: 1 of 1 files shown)


content/rancher/v2.x/en/tools/pipelines/configurations/_index.md
```diff
@@ -474,6 +475,48 @@ branch:
...
+
+{{% tabs %}}
+{{% tab "By UI" %}}
+1. From the context menu, open the project for which you've configured a pipeline. Then select the **Pipelines** tab.
+
+1. From the pipeline for which you want to edit the timeout, select **Ellipsis (...) > Edit Config**.
+
+1. Enable the **Notification** option.
+
+1. Select conditions for notifications. For example if you want to receive notifications when on execution failure, select **Failed**
```
MBishop17:
```suggestion
1. Select conditions for notifications. For example, if you want to receive notifications when on execution failure, select **Failed** . 
```

Nitpicks.

content/rancher/v2.x/en/tools/pipelines/configurations/_index.md
```diff
@@ -474,6 +475,48 @@ branch:
...
+
+1. From the pipeline for which you want to edit the timeout, select **Ellipsis (...) > Edit Config**.
+
+1. Enable the **Notification** option.
+
+1. Select conditions for notifications. For example if you want to receive notifications when on execution failure, select **Failed**  
+
+1. If you don't have any existing notifier, click **create one** and you will be directed to the notifiers page. Follow the [instruction](https://rancher.com/docs/rancher/v2.x/en/tools/notifiers-and-alerts/#adding-notifiers) to add a notifier. If you have already configured notifiers, click **Add Recipient** button.
+
+1. Select a notifier from the dropdown, then input recipient or use the default one. By clicking **Add Recipient** button, you can configure to send notifications to different notifiers or multiple recipients.
```
MBishop17:
```suggestion
1. Select a notifier from the dropdown, then input recipient or use the default one. By clicking the **Add Recipient** button, you can send notifications to different notifiers or multiple recipients.
```

MBishop17:
@gitlawr and @deniseschannon, I suggested a couple of minor grammar changes, but it lgtm otherwise.

gitlawr:
@MBishop17 Thanks, your suggestions are committed.

gitlawr:
Related to issue: https://github.com/rancher/rancher/issues/16827

loganhz:
LGTM

loganhz:
Please add `_Available as of v2.2.0_` so that people know when these features were available.

https://github.com/rancher/docs/pull/1199/files#diff-8ad9c71b76b83c64e18dda58fa1e8c68R9

content/rancher/v2.x/en/tools/pipelines/configurations/_index.md
```diff
@@ -474,6 +475,48 @@ branch:
...
+  - name: Build something
+    steps:
+    - runScriptConfig:
+        image: busybox
+        shellScript: ls
+notification:
+  recipients:
+  - recipient: "#mychannel"
+    # notifier name or id
+    notifier: "my-slack-notifier"
```
loganhz: name is not unique. Can you check the code? It looks like it's a bug

deniseschannon:
@gitlawr I've updated your PR in the Github desktop so please don't forcepush over this branch.

content/rancher/v2.x/en/tools/pipelines/configurations/_index.md
gitlawr: This is intentional. `name` is easier for users who would write configurations by YAML. We don't use the name when users configure it via the UI. And When multiple targets match, one of them will be picked.

deniseschannon: But how do you pick? This is poor UX as you're assuming that users will have correctly named their notifiers with different names. if we expect to have unique names in a cluster, then we should change it so that notifier names need to be unique. 

We cannot just randomly pick one.

gitlawr: Then we should remove the name option and use only IDs in the docs

deniseschannon: @gitlawr can you confirm the example ID is what the API is expecting?

@loganhz we should file a bug for the name. we should error out for notifications if we try to use a name. cc: @cloudnautique to confirm that this behavior makes sense

loganhz: Filed the issue https://github.com/rancher/rancher/issues/18726

gitlawr: Need to change the example ID to `c-dbcqx:n-sd2rr`. This is how a notifier ID looks like

loganhz: Yes, Denise already changed it to the correct one.

PR #1045, opened 2018-12-06
adding note that Rancher load balancer shouldn't serve double duty wi…

MBishop17:
…th for other apps


alena1108:
https://github.com/rancher/rancher/issues/16301

MBishop17:
@deniseschannon and @alena1108 , is this text more accurate?

>**Using Rancher in a High Availability Configuration?**
>
>Refrain from adding an Ingress to the `local` cluster. The default Nginx Ingress Controller that Rancher uses acts as a global load balancer for _all_ clusters managed by Rancher, including the `local` cluster.  Therefore, when users try to access an application, your Rancher connection may drop due to the Nginx configuration being reloaded. 
>
>If you need to use an Ingress with your `local` cluster, we recommend changing to a different Ingress Controller that reloads more seamlessly, like [HAProxy](http://www.haproxy.org/) or [Traefik](https://traefik.io/).

alena1108:
@MBishop17 @deniseschannon 

> If you need to use an Ingress with your local cluster, we recommend changing to a different Ingress Controller that reloads more seamlessly, like HAProxy or Traefik.

I think we should be careful when make this recommendation as we only have nginx ingress controller as a part of the server chart covered by our SLA

MBishop17:
> @MBishop17 @deniseschannon
> 
> > If you need to use an Ingress with your local cluster, we recommend changing to a different Ingress Controller that reloads more seamlessly, like HAProxy or Traefik.
> 
> I think we should be careful when make this recommendation as we only have nginx ingress controller as a part of the server chart covered by our SLA

@alena1108, does that mean that there currently isn't a way to change your ingress controller? Or just that we shouldn't call out Nginx competitors by name and keep it general.

deniseschannon:
We're saying you can only use the NGINX ingress controller to service Rancher and no other apps.

PR #1056, opened 2018-12-10
Add note on Helm v2.12.0 and cert-manager

superseb:
https://github.com/rancher/rancher/issues/17002


deniseschannon:
@superseb not sure if we need to add in the GH issue into the docs.

superseb:
Updated

PR #1064, opened 2018-12-17
Add more info on importing clusters

superseb:


deniseschannon:
Do you think adding the notes of how to fix an API endpoint for imported clusters should be added? https://github.com/rancher/rancher/issues/15322

superseb:
I would merge this to have correct info for importing clusters and have a docs issue depending on how the referenced issue is going to be solved.

PR #1073, opened 2018-12-26
Suggest using --reuse-values during upgrade

bradjones1:
...instead of manually copying them over.


superseb:
Is this tested with all installation scenarios? Behavior of this flag has been inconsistent.

bradjones1:
@superseb That I couldn't say; it has been fine for me and is definitely more straightforward than exporting and copying into the upgrade command. There is https://github.com/helm/helm/issues/2948 which seems to address some more advanced functionality than I think is present in the Rancher chart? Perhaps someone more familiar with the Helm internals can chime in.

deniseschannon:
Based on the helm issue that was pointed out, I'm not going to include this change to the docs.

bashofmann:
@deniseschannon Since the issue has been fixed in Helm end of last year, would it make sense to add this change to the docs now?
Using `--reuse-values` is significantly easier then the currently documented approach. I tested an upgrade today vom 2.3.6 to 2.4.2 and it worked fine.

PR #1082, opened 2018-12-30
minor update to k8s resources page

aemneina:
updated example kubeconfig filename to reflect what rke would spit out, hopefully making it consistent with our other docs.


superseb:
The Rancher docs regarding RKE say to use `rancher-cluster.yml` (https://rancher.com/docs/rancher/v2.x/en/installation/ha/kubernetes-rke/#create-the-rancher-cluster-yml-file) and it results in `kube_config_rancher-cluster.yml` (https://rancher.com/docs/rancher/v2.x/en/installation/ha/kubernetes-rke/#testing-your-cluster). If there is a reference to `kube_config_cluster.yml` for Rancher HA somewhere, it should be corrected there.

PR #1090, opened 2019-01-07
RKE - kube-api: always_pull_images option

jgreat:
Docs for always_pull_images option.

https://github.com/rancher/rke/pull/1059


alena1108:
@deniseschannon it will be in rke v2.0.0

LGTM

PR #1119, opened 2019-01-30
simplify Helm upgrade

tobru:
Helm upgrade supports the `--reuse-values` parameter which makes it much easier to upgrade. It also helps to prevent mistakes by manually getting values from the prior installation.

```
      --reuse-values             when upgrading, reuse the last release's values and merge in any overrides from the command line via --set and -f. If '--reset-values' is specified, this is ignored.
```


superseb:
Thanks, this will be tested in https://github.com/rancher/docs/pull/1073

PR #1139, opened 2019-02-12
Update _index.md

acaiblog:


deniseschannon:
@acaiblog thanks for the PR, but we've always used 17.03.2 to validate/support Rancher.

PR #1151, opened 2019-02-15
Added missing "=" while creating cluster-admin.

vishnuvpotti:
The "=" was missing in this snip, after --clusterrole.  '--clusterrole cluster-admin \'
When the command is executed without = , the ntillre won't function properly. 
Refer https://stackoverflow.com/questions/54683563/helm-error-error-the-server-has-asked-for-the-client-to-provide-credentials


deniseschannon:
Thanks @vishnuvpotti !

PR #1152, opened 2019-02-15
Restoring downstream kubernetes cluster

galal-hussein:
https://github.com/rancher/rancher/issues/17867

### DIFF
--- a/content/rancher/v2.x/en/backups/restorations/restoring-downstream/_index.md
+++ b/content/rancher/v2.x/en/backups/restorations/restoring-downstream/_index.md
@@ -1,0 +1,34 @@
+---
+title: Restoring Downstream Kubernetes Cluster
+weight: 370
+aliases:
+  - /rancher/v2.x/en/installation/after-installation/restoring-downstream-cluster/
+---
+
+Before version 2.2.x Rancher doesn't support backup and restoration of etcd snapshots, to be able to restore etcd cluster on a downstream kubenretes cluster you can run the following procedure:
+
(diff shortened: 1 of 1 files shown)


deniseschannon:
@galal-hussein I want to think more about this and where this information should go. Since it's about managing the user clusters (not the cluster running Rancher, I'm not sure where you've put it is the best place). I'll let you know where I think it should go when I've figured out where.

content/rancher/v2.x/en/backups/restorations/restoring-downstream/_index.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Restoring Downstream Kubernetes Cluster
+weight: 370
+aliases:
+  - /rancher/v2.x/en/installation/after-installation/restoring-downstream-cluster/
+---
+
+Before version 2.2.x Rancher doesn't support backup and restoration of etcd snapshots, to be able to restore etcd cluster on a downstream kubenretes cluster you can run the following procedure:
+
+1- If etcd lose quorum then the Kubernetes cluster will report a failure in Rancher, the first thing to do is to remove all other etcd nodes and leave only one etcd node in the cluster.
```
aemneina: how does one know which node to keep, and which to discard. should they back anything up before removing etcd nodes. What does removing look like, is it a node delete?

ankur111: we should provide the commands to remove all but the node you want to keep.

deniseschannon:
@galal-hussein can you update this PR?

deniseschannon:
Merging as is as @superseb is going to fix it up and move it to the correct location

PR #1165, opened 2019-02-26
Update _index.md

meappy:
Source should be sg-xxx (rancher-nodes) not 30000-32767?


vincent99:
No? NodePorts are a port you're excitedly requesting to be published to the outside world.   And there is no reason for nodes to use them to talk to one another, the same port is mapped to the same service on every node.

PR #1202, opened 2019-03-07
suggestion doc change

mjschmidt:
I wasn't able to get my nginx conf to work without adding an empty events and wrapping the whole request with http


superseb:
What NGINX version was used?

mjschmidt:
I think 1.14 I need to double check that, but I was following rancher docs when I did this.

mjschmidt:
So those may need to be updated as well on the ha install?

mjschmidt:
it is 1.14

superseb:
Thanks for the report, I've made changes for both NGINX configs here https://github.com/rancher/docs/pull/1229.

Issue was that the config is made as a "vhost" config, not the main config. As the L4 example is also a full config, I've synced both locations and made them full config as well.

PR #1205, opened 2019-03-07
Clarify availability of DNS provider options

superseb:


deniseschannon:
in most other docs, we've done just <major number>.<minor number>. let's keep that consistent

so this should be `_Available as of v0.2_`

do you think we should write RKE in front of it so people know?

superseb:
Putting RKE in front would not be consistent, or we have to change it everywhere. It now says `v0.2`

superseb:
And now it says `v0.2.0`

PR #1270, opened 2019-03-20
Fix typo and add restarting pods info

galal-hussein:


alena1108:
@galal-hussein do you think it's worth mentioning that certificates generated by rke on rancher 2.2, has 10 years of expiration vs previously been 1 year?

deniseschannon:
@alena1108 i removed the reference of how long the cert is for.

PR #1274, opened 2019-03-20
Removed extra column in separator row of table

boxboatjeremy:


superseb:
Thanks for the PR, this should be fixed in staging already and will be merged to master soon

PR #1309, opened 2019-03-25
Removed Contact Support

jambajaar:
Removed this --> Please contact Rancher support if your worker nodes on AWS, virtualization clusters, or bare metal servers don't belong to the same layer 2 network.


deniseschannon:
Fixed in . https://github.com/rancher/docs/pull/1330. It will go live tomorrow.

PR #1312, opened 2019-03-25
Pipeline changes

gitlawr:
Changes include:   
1. Repo list is based on the current user, not the project owner that enables project pipeline.
2. Branch selection(both UI/yaml) is to determine where the pipeline file locates, and is different from the branch conditions in pipeline files.
3. Add missing event trigger configuration for a repository.
4. Secrets can be in the pipeline namespace so that the distribution scope is smaller.
5. A manual run is done by UI, the Git CLI way is included in the webhook triggered runs.
6. Change the word `admin permission` to more generic `webhook management permission`, since `admin` is specific to Github.

### DIFF
--- a/content/rancher/v2.x/en/k8s-in-rancher/pipelines/_index.md
+++ b/content/rancher/v2.x/en/k8s-in-rancher/pipelines/_index.md
@@ -45,7 +45,7 @@ After the version control provider is authorized, you are automatically re-direc
 
 1. Click on **Configure Repositories**.
 
-1. A list of repositories are displayed. The list of repositories displayed are based on the [user that set up the version control provider]({{< baseurl >}}/rancher/v2.x/en/project-admin/tools/pipelines/#version-control-providers).
+1. A list of repositories are displayed. If you are configuring repositories the first time, click on **Authorize & Fetch Your Own Repositories** to fetch your repository list.
 
 1. For each repository that you want to set up a pipeline, click on **Enable**.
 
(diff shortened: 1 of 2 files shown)


content/rancher/v2.x/en/k8s-in-rancher/pipelines/_index.md
```diff
@@ -68,36 +68,7 @@ Now that repositories are added to your project, you can start configuring the p
...
-# branch conditions for the pipeline
-branch:
-  include: [ master, feature/*]
-  exclude: [ dev ]
-```
-<br>
-{{% /tab %}}
-{{% /tabs %}}
-    {{% /accordion %}}
+1. Select which `branch` to use from the list of branches.
```
deniseschannon: Why did you remove this section? How does one specify what branch in the yaml file?

content/rancher/v2.x/en/k8s-in-rancher/pipelines/_index.md
```diff
@@ -527,6 +498,18 @@ Within a pipeline, there are multiple advanced options for different parts of th
...
+1. From the **Global** view, navigate to the project that you want to configure pipelines.
+
+1. Select **Workloads** in the navigation bar and then select the Pipelines tab.
+
+1. From the repository for which you want to change the event triggers, select **Ellipsis (...) > Setting**.
+
+1. Click on the check box to enable/disable triggering for **Push**, **Pull Request** or **Tag** event.
+
+1. Click **Save**.
+
```
deniseschannon: This exists as a pipeline trigger in the tabs so why is it here?

There is no "setting" option in the ellipsis.

content/rancher/v2.x/en/k8s-in-rancher/pipelines/_index.md
gitlawr: When a user update the pipeline config via either UI or Yaml, he needs to pick a branch. This indeed means in which branch he is editing the pipeline file.

The branch condition in the pipeline yaml is described in the `Trigger Rules` section and is different in the concept.

content/rancher/v2.x/en/k8s-in-rancher/pipelines/_index.md
gitlawr: When you do edit-config, all the trigger rules are the configs in the pipeline file. and we don't have event type for a pipeline trigger. The reason for not configuring events in that way is that users should not be able to activate pull request events by submitting a PR and change the pipeline file in it. This should be something controlled on the management side. So this is a repository setting to configure allowable events.

There is no `setting` for the example repos because users don't manage webhooks for them. It is present when it's a user configured repo.

content/rancher/v2.x/en/k8s-in-rancher/pipelines/_index.md
deniseschannon: Okay, I missed the dropdown in the view/edit yaml. thanks!

content/rancher/v2.x/en/k8s-in-rancher/pipelines/_index.md
deniseschannon: Cool, i see it in the user repos!

deniseschannon:
LGTM

PR #1344, opened 2019-03-28
Updating nginx information

mjschmidt:
Based on functional testing and usage. Happy to update more, but let me know if this is how you want this documented.

### DIFF
--- a/content/rancher/v2.x/en/installation/ha/create-nodes-lb/nginx/_index.md
+++ b/content/rancher/v2.x/en/installation/ha/create-nodes-lb/nginx/_index.md
@@ -72,8 +72,11 @@ After installing NGINX, you need to update the NGINX configuration file, `nginx.
 Instead of installing NGINX as a package on the operating system, you can rather run it as a Docker container. Save the edited **Example NGINX config** as `/etc/nginx.conf` and run the following command to launch the NGINX container:
 
 ```
-docker run -d --restart=unless-stopped \
+docker run -d --restart=unless-stopped --name <rancher.my.org> \
   -p 80:80 -p 443:443 \
   -v /etc/nginx.conf:/etc/nginx/nginx.conf \
   nginx:1.14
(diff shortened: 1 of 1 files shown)


superseb:
I don't see how a container name can change the behavior, this probably needs some more context before we can change anything. Are you running it inside docker-compose or some other way? 

I'd like to know what is wrong before changing anything, can you supply exact steps to reproduce the setup where Rancher doesn't handle the requests (and can you also elaborate what that means exactly?)

mjschmidt:
@superseb when you run nginx in a container the "hostname" that request is coming from becomes the container id rather than the machine name. That is why you need to set the container hostname to the same thing as the machine you are coming from. When you launch rancher in helm, you set the address that rancher expects the requests to come from. If the requests come from anywhere else but the host that it expects the requests from you get a 404 backend.


To reproduce:
1) Get 4 vms or aws instances or 4 physical machines and load docker onto them per the rancher instructions.
2) run the nginx load balancer in a container on a machine and point that nginx at the other 3 machines
3) deploy the kubernetes cluster with rke to the other 3 machines
4) helm install cert manager and rancher per the instructions, and set the expected hostname to the name of the machine
```
helm install rancher-stable/rancher \
  --name rancher \
  --namespace cattle-system \
  --set hostname=machine1
```

In the browser navigate to machine1's address
This should fail
docker stop and docker rm the nginx container on machine1
run the same docker container, but this time set the hostname of the container to machine1

Does that provide more detail/context?

mjschmidt:
@superseb any thoughts?

mjschmidt:
@superseb I am good with the changes you made. Do I need to actually go off and implement them and submit a follow on pull request?

superseb:
This was not reproduced and not hit by anyone, if you have a set of instructions to reproduce this behavior in an automated fashion on a cloud, I would love to run it and debug it.

PR #1353, opened 2019-04-02
Fix external link for Kubernetes webhook authentication

janeczku:
The current docs for the `webhook` authentication strategy have a link to upstream docs on webhook authorisation instead of webhook authentication.


erikwilson:
LGTM!

PR #1378, opened 2019-04-12
'Use Catalog Templates', not 'User Catalog Templates'

stefanlasiewski:
Pretty sure this is a typo. There's a big difference between 'Use Catalog Templates' and 'User Catalog Templates', so it would be good for someone to validate this.

Here's a snapshot from the UI, which says 'Use Catalog Templates':

image


deniseschannon:
Thanks @stefanlasiewski!

PR #1390, opened 2019-04-18
Update the example to backup to use the correct tag

shawncarr:
The documentation states to use `<RANCHER_CONTAINER_TAG>` but the example has `rancher-data` as the container tag which would result most likely in something like `docker: Error response from daemon: No such container: rancher-data.`. 

So replaced `rancher-data` with `<RANCHER_CONTAINER_TAG>` to make consistent with the existing documentation.


shawncarr:
@dnoland1 you are correct. I misread the `Before You Start`.

PR #1405, opened 2019-04-29
additional detail to rancher ha (back/restore) and etcd snapshots

ChrisMcKee:
Hopefully this will add some value and also help emphasise the improvements gained via RKE2 and Rancher 2.2s snapshotting model.


deniseschannon:
Thanks @ChrisMcKee!

PR #1408, opened 2019-04-30
Replacing cert only requires same CA if private CA signed cert is being used

amherrington13:


superseb:
* Can you squash the commits into 1?
* Can you replace supported with possible (not related to the change but supported needs to go out of the docs anyway)
* I assume this is a tested scenario?

amherrington13:
I've squashed the commits into 1 and updated the wording to "possible". 
Yes, this scenario has been tested.

PR #1411, opened 2019-05-06
server-chart options for ingress.configurationSnippet and antiAffinity rules

jgreat:
- `ingress.configurationSnippet` should be part to the 2.2.3 release and backported to latest 2.1/2.0 releases/
- `antiAffinity`  option was added in a while ago, just forgot to put it into the docs.


content/rancher/v2.x/en/installation/ha/helm-rancher/chart-options/_index.md
```diff
@@ -98,7 +100,13 @@ To customize or use a different ingress with Rancher server you can set your own
 Example on setting a custom certificate issuer:
 
 ```plain
- --set ingress.extraAnnotations.'certmanager\.k8s\.io/cluster-issuer'=ca-key-pair
+--set ingress.extraAnnotations.'certmanager\.k8s\.io/cluster-issuer'=ca-key-pair
+```
+
+Example on setting a static proxy header with `ingress.configurationSnippet`. This value is parsed like a template so variables can be used.
```
deniseschannon: @jgreat do you think we need to say it's available starting specific versions?

PR #1427, opened 2019-05-12
Correct version references for certificate rotation.

jannylund:


deniseschannon:
Thanks this is already fixed in a different PR.

PR #1440, opened 2019-05-20
Update windows feature notes

loganhz:


loganhz:
FYI. Windows won't work in 2.2.3 and it is not hidden in 2.2.3 release.

Issue #1389, opened 2019-04-17
Documentation for "Nodes > Draining a Node" is out of date and looks different then Rancher 2.2


stefanlasiewski:
I'm trying to understand the details of draining a node, as explained at https://rancher.com/docs/rancher/v2.x/en/cluster-admin/nodes/#draining-a-node.

The screenshot and the complementary text is fairly different then what I see in Rancher 2.2.2. 

For example, here's a screenshot from Rancher 2.2.2:

![image](https://user-images.githubusercontent.com/142752/56323474-da403700-6120-11e9-8d6d-7d6dea551784.png)

The current screen is substantially different than the one used in the documents:

![image](https://user-images.githubusercontent.com/142752/56323522-ffcd4080-6120-11e9-858e-70acc8fb8654.png)

The document explains options that aren't present in 2.2.2, and I'm not sure if these map to the new options:

> * Even if there are pods not managed by a ReplicationController, ReplicaSet, Job, DaemonSet or StatefulSet
> * Even if there are DaemonSet-managed pods

vincent99:
There are [5 options](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#drain), but most combinations of the 3 flags are fairly useless so they were combined into one choice.

"Safe" == ignoreDaemonSets  (same as the 2.1 screen default)

"Aggressive" == deleteLocalData, force, ignoreDaemonSets

And the timeouts have controls for the magic values instead of explaining that you can put in `-1`/`0`.

PR #1448, opened 2019-05-22
Remove outdated references in node drain section

catherineluse:
This addresses issue #1389, "Documentation for "Nodes > Draining a Node" is out of date and looks different then Rancher 2.2."

The screen shot has been updated and the options reflect what is in the Rancher UI.


deniseschannon:
@catherineluse can we remove the screenshot in the file? 

also, can we save the previous docs and separate what are docs for v2.2.x and prior to v2.2.x?

PR #1450, opened 2019-05-23
Make headers look better in docs for draining a node

catherineluse:


deniseschannon:
I think this PR is old as it's updated okay on the site.

PR #1466, opened 2019-05-31
Update for RancherOS v1.5.2

niusmallnan:


content/os/v1.x/en/installation/running-rancheros/workstation/boot-from-iso/_index.md
```diff
@@ -5,7 +5,7 @@ weight: 102
 
 The RancherOS ISO file can be used to create a fresh RancherOS install on KVM, VMware, VirtualBox, or bare metal servers. You can download the `rancheros.iso` file from our [releases page](https://github.com/rancher/os/releases/).
 
-You must boot with at least **1280MB** of memory. If you boot with the ISO, you will automatically be logged in as the `rancher` user. Only the ISO is set to use autologin by default. If you run from a cloud or install to disk, SSH keys or a password of your choice is expected to be used.
```
deniseschannon: I think we need a chart of which version requires what amount of memory cause changing it here won't work if I'm still running on v1.5.1.

niusmallnan: @deniseschannon I set up a link here, which shows a chart about the memory requirements.

PR #1468, opened 2019-06-04
Add docs for Global Registry

gitlawr:


catherineluse:
Questions:

About the configuration options table:

- I noticed the first two rows on the table - the admin password an encryption key - aren't on the UI. Do you use these only when you set up Global Registry for the first time?
- I noticed the UI has a red asterisk for required fields. Should the docs indicate which fields are required as well?
- Should the configuration options table be broken into separate tables (Registry, Database, Redis, Clair, Notary) so that each one can have its own brief overview/explanation?
- Would you specify options for both Database and Redis? Or just one or the other?
- Should the configuration options table indicate which options only apply to when you are using an internal type of database, and which options apply for an external database?
- Should there be some examples for how to do the configuration options? Maybe an example for internal database, and an example for an external database
- When you specify options for Redis, Clair, and Notary, how much do these options affect? everything in the global registry? or just a scheduled workload on selected nodes?
- Which options are only used when you set up the registry for the first time?

 
About the "Using the Global Registry" section:

- It says you can “use Global Registry as a private registry in Rancher projects.” is this optional? If so, what’s the advantage?
- This section links to a document called How to Use Registries (https://rancher.com/docs/rancher/v2.x/en/k8s-in-rancher/registries/) which defines a Registry as a secret, which might be confusing to some users. Is it ok to say “Deployments can authenticate with the Global Registry and then pull a Docker image hosted on it.”?

 
About the UI:

- I see that the UI has Docker commands for pulling and pushing Docker images to the registry. Should those be in the docs as well?
- I see the UI, under each set of configuration options, has a button for "Select the nodes where Registry workloads will be scheduled to" and it lets you select a key-value pair. What do you use for the key and value? Should there be an example in the docs? (I tried the docker login command and got "Error response from daemon: Get https://sowmyaha.qa.rancher.space/v2/: x509: certificate signed by unknown authority")
- What's happening when you schedule a registry workload to one of the selected nodes? Why is there a separate selector button for each set of options?

loganhz:
Close in favor of https://github.com/rancher/docs/pull/1490

PR #1472, opened 2019-06-05
add alert prereq: enable cluster monitoring

lalyos:


deniseschannon:
This is not a pre-req for alerting.

PR #1490, opened 2019-06-07
Add docs for Global Registry

loganhz:


loganhz:
Questions:

About the configuration options table:

- I noticed the first two rows on the table - the admin password an encryption key - aren't on the UI. Do you use these only when you set up Global Registry for the first time?
- I noticed the UI has a red asterisk for required fields. Should the docs indicate which fields are required as well?
- Should the configuration options table be broken into separate tables (Registry, Database, Redis, Clair, Notary) so that each one can have its own brief overview/explanation?
- Would you specify options for both Database and Redis? Or just one or the other?
- Should the configuration options table indicate which options only apply to when you are using an internal type of database, and which options apply for an external database?
- Should there be some examples for how to do the configuration options? Maybe an example for internal database, and an example for an external database
- When you specify options for Redis, Clair, and Notary, how much do these options affect? everything in the global registry? or just a scheduled workload on selected nodes?
- Which options are only used when you set up the registry for the first time?

 
About the "Using the Global Registry" section:

- It says you can “use Global Registry as a private registry in Rancher projects.” is this optional? If so, what’s the advantage?
- This section links to a document called How to Use Registries (https://rancher.com/docs/rancher/v2.x/en/k8s-in-rancher/registries/) which defines a Registry as a secret, which might be confusing to some users. Is it ok to say “Deployments can authenticate with the Global Registry and then pull a Docker image hosted on it.”?

 
About the UI:

- I see that the UI has Docker commands for pulling and pushing Docker images to the registry. Should those be in the docs as well?
- I see the UI, under each set of configuration options, has a button for "Select the nodes where Registry workloads will be scheduled to" and it lets you select a key-value pair. What do you use for the key and value? Should there be an example in the docs? (I tried the docker login command and got "Error response from daemon: Get https://sowmyaha.qa.rancher.space/v2/: x509: certificate signed by unknown authority")
- What's happening when you schedule a registry workload to one of the selected nodes? Why is there a separate selector button for each set of options?

loganhz:
@catherineluse Copied the docs from the other PR. So let's use this PR for global registry

loganhz:
Questions

### About the configuration options table:

> I noticed the first two rows on the table - the admin password an encryption key - aren't on the UI. Do you use these only when you set up Global Registry for the first time?

admin password is the initial password for Harbor admin user. After we enabled Global reigstry, admin user needs to go Harbor UI to change the password for admin user.
So even users change it in the table. It won't take effect. So we will hide the admin password field after enabled global registry.

For encryption key, it's a bug. We should always show it.

> I noticed the UI has a red asterisk for required fields. Should the docs indicate which fields are required as well?

added the required column

> Should the configuration options table be broken into separate tables (Registry, Database, Redis, Clair, Notary) so that each one can have its own brief overview/explanation?

Done

>  Would you specify options for both Database and Redis? Or just one or the other?

Both. Database is for Harbor metadata and other Harbor data. Redis is a cache layer. So we need to config both of them.

> Should the configuration options table indicate which options only apply to when you are using an internal type of database, and which options apply for an external database?

Done. But I put the info in required column...

> Should there be some examples for how to do the configuration options? Maybe an example for internal database, and an example for an external database

That's a good point. We can provide the example options. Could you add the example configs to somewhere? I don't know where should I put.

> When you specify options for Redis, Clair, and Notary, how much do these options affect? everything in the global registry? or just a scheduled workload on selected nodes?

each option maps to a field of a workload, a config map or a secret. When you change an option, it will only affect the workload which is using that option.

> Which options are only used when you set up the registry for the first time?

Everything is used for first time. But after enabled, admin password will be hidden. 
And we don't allow users to switch Database and Redis between internal and external
Also, we don't allow users to change the backend type for registry between s3 and filesystem



### About the "Using the Global Registry" section:

> It says you can “use Global Registry as a private registry in Rancher projects.” is this optional? If so, what’s the advantage?

Users can push images to Global registry and when users want to deploy the workloads in a project. Users can use the images inside the global registry. Before users can pull images from a private registry. Users have to configure the registry credential in Rancher UI for the private registry. 

> This section links to a document called How to Use Registries (https://rancher.com/docs/rancher/v2.x/en/k8s-in-rancher/registries/) which defines a Registry as a secret, which might be confusing to some users. Is it ok to say “Deployments can authenticate with the Global Registry and then pull a Docker image hosted on it.”?

I am not sure. I think it tries to explain how registry credential works in Rancher

### About the UI:

> I see that the UI has Docker commands for pulling and pushing Docker images to the registry. Should those be in the docs as well?

No, I think. 

> I see the UI, under each set of configuration options, has a button for "Select the nodes where Registry workloads will be scheduled to" and it lets you select a key-value pair. What do you use for the key and value? Should there be an example in the docs? (I tried the docker login command and got "Error response from daemon: Get https://sowmyaha.qa.rancher.space/v2/: x509: certificate signed by unknown authority")

You can add  key/value pair as labels to nodes. So you can add node selector to match the labels which you added in a node. Then the workload will be deployed to that particular node.

> What's happening when you schedule a registry workload to one of the selected nodes? Why is there a separate selector button for each set of options?

Users might want to run different components in different dedicated nodes.

catherineluse:
Logan,

Thanks for the detailed answers. This clarifies a lot.

I would like to make some changes to this commit, can you please allow edits from maintainers? https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

To answer your question about where the example configuration commands should go, I no longer think it's necessary because of the new clarifications in the configuration options table.

According to my understanding, I think the workflow for setting up the registry works like this. Please correct me if I'm wrong:

1. Create separate databases for Harbor registry, Harbor data, and Redis
2. Optional: Create databases for Notary server and Clair
3. Set up Harbor credentials
4. Enable global registry in Rancher UI
5. Use Rancher UI to specify which components of the global registry run in different dedicated nodes
6. Access Docker images through the Harbor UI or through global registry hostname

Is that the basic workflow?

PR #1505, opened 2019-06-12
New global registry

catherineluse:


deniseschannon:
Closing this for now as we aren't adding in global registry for now. @catherineluse you should keep this branch around in case we end up adding it back in.

PR #1506, opened 2019-06-13
Add warning about running HA Rancher on hosted K8s

ebauman:
A warning should be added to the top of the HA install docs that this is not supported nor recommended on hosted Kubernetes clusters.
I am unsure if the syntax of the warning is correct - if not, I can change it to something appropriate. However, I believe this deserves to be a **warning** instead of just a note.


ebauman:
Seeing that CI is failing (Hugo, specifically) with an error about my warning tags, I should use a different syntax. If someone can recommend, I will update this PR.

deniseschannon:
I'm not sure if that tag is implemented with the new site. Can you just make it another `> **Note**`?

I know it's not ideal as there is an `important` section right underneath but that would be the quickest way to get it out.

ebauman:
Changed to use > **Important:** notation.
Apologies for delay.

PR #1522, opened 2019-06-19
Menu clarification on Quickstart Workload pages

cbron:
URLs this PR updates: 
* https://rancher.com/docs/rancher/v2.x/en/quick-start-guide/workload/quickstart-deploy-workload-ingress/
* https://rancher.com/docs/rancher/v2.x/en/quick-start-guide/workload/quickstart-deploy-workload-nodeport/

If you had already clicked around in the cluster before you started this guide, when you do step 3 you will default to the last page you had open. That might be inside the Apps menu item, or Workloads->LB. We should clarify that they want to be on Workloads->Workloads, even though its a bit redundant.


catherineluse:
Looks good to me

PR #1529, opened 2019-06-21
Create doc for prometheus metrics usage

orangedeng:
Relate issue:
https://github.com/rancher/rancher/issues/20864

### DIFF
--- a/content/rancher/v2.x/en/cluster-admin/tools/monitoring/custom-metrics/_index.md
+++ b/content/rancher/v2.x/en/cluster-admin/tools/monitoring/custom-metrics/_index.md
@@ -1,0 +1,489 @@
+---
+title: Prometheus Custom Metrics Adapter
+weight: 5
+---
+
+After you've enabled [cluster level monitoring]({{< baseurl >}}/rancher/v2.x/en/cluster-admin/tools/monitoring/#enabling-cluster-monitoring), You can view the metrics data from Rancher. You can also deploy the Prometheus custom metrics adapter then you can use the HPA with metrics stored in cluster monitoring.
+
+## Deploy Prometheus Custom Metrics Adapter
+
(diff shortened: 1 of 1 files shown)


catherineluse:
Hi,
There's a lot of good information in here. I have a few questions about this document, mainly oriented toward adding a little context and making it easier for someone who is new to Prometheus:

- We already have some instructions about installing k8s-prometheus-adapter at this link: https://rancher.com/docs/rancher/v2.x/en/k8s-in-rancher/horitzontal-pod-autoscaler/manage-hpa-with-kubectl/#configuring-hpa-to-scale-using-custom-metrics-with-prometheus Should this document link to the other one or be combined into the same document? Or are they used in different situations? Why doesn’t this PR have kubectl commands like the instructions in the linked doc?
- What are the prerequisites for setting up the adapter? Is the only prerequisite that you have to enable cluster level monitoring?
- What filenames and directory locations do these config files have? I found it hard to follow these instructions because I had difficulty figuring out which of these code snippets go together in the same file and which go in different files.
- The URL for sample-config.yml doesn’t work for me. Can you validate that it’s the right URL?
- One of the code samples mentions “cAdvisor metrics.” What are these metrics? How does cAdvisor work with Prometheus?
- This document seems to assume knowledge of Kubernetes, for example it assumes that people know the following Kubernetes concepts of HPA, service accounts, granting permissions to service accounts, Rolebinding and ClusterRoleBinding. I can find some links to the Kubernetes docs for these, but do you know of any other good resources to link for people who might need extra help with these topics?
- I found the section on Discovery, Association, Naming, and Querying to be somewhat complicated and I had difficulty understanding the business purpose behind the examples. I think it's confusing to me because it's organized more by feature than by the user goals (setting up HPA in various ways).  Do you have any links to resources that provide more of a tutorial or walkthrough for how you would translate certain metrics into Prometheus configurations?

Issue #1413, opened 2019-05-07
Fix documentation for usage / behaviour of private registries in projects


janeczku:
[This page](https://rancher.com/docs/rancher/v2.x/en/k8s-in-rancher/registries/) talks about configuring a private registry in a project and concludes with:

> For more information on adding a registry to a workload, see Deploying Workloads.

But when navigating to the [linked page](https://rancher.com/docs/rancher/v2.x/en/k8s-in-rancher/workloads/deploy-workloads/) there is no information at all with regards to adding registries to workloads.

We should fix this and also document the general behaviour of the private registry settings and its limitations (only works for workloads deployed in GUI, etc.).

PR #1548, opened 2019-07-01
Add detail about deploying workloads using a private registry with Rancher UI and kubectl

catherineluse:
This PR addresses this issue https://github.com/rancher/docs/issues/1413


content/rancher/v2.x/en/k8s-in-rancher/registries/_index.md
```diff
@@ -21,14 +27,83 @@ Currently, credentials are pulled automatically only if the workload is created
...
+1. Click **Deploy.**
+1. Enter a unique name for the workload and choose a namespace.
+1. In the **Docker Image** field, enter the URL of the path to the Docker image in your private registry. For example, if your private registry is on Quay.io, you could use `quay.io/<Quay profile name>/<Image name>`.
+1. Click **Launch.**
+
+**Result:** Your deployment should launch, authenticate using the private registry credentials you added in the Rancher UI, and pull the Docker image that you specified.
+
+### Using the Private Registry with kubectl
+
+When you create the workload using `kubectl`, you need to configure the pod so that its YAML has the path to the private registry. You also have to specify the registry secret because the pod only automatically gets access to the private registry credentials if it is created in the Rancher UI.
```
alena1108: I think we should re-write this paragraph a little. `so that its YAML has the path to the private registry` >> we should elaborate that the image has the path to the private registry.
`You also have to specify the registry secret` - we have to a) create and b) reference the secret in the pod specification.

We also use YAML and .yml in diff places; we should stick to one way for consistency.

content/rancher/v2.x/en/k8s-in-rancher/registries/_index.md
```diff
@@ -21,14 +27,83 @@ Currently, credentials are pulled automatically only if the workload is created
...
+  name: private-reg
+spec:
+  containers:
+  - name: private-reg-container
+    image: quay.io/<Quay profile name>/<image name>
+  imagePullSecrets:
+  - name: testquay
+```
+
+You can use `kubectl` to create the secret with the private registry credentials. In this example, the secret is named `testquay`:
```
alena1108: We should mention that the secret has to be created in the same namespace where the workload gets deployed.

alena1108: And mention that the example in the doc is for the default namespace.

alena1108:
LGTM

PR #1556, opened 2019-07-03
self-assessment and hardening guide for 2.2

drpebcak:


content/rancher/v2.x/en/security/benchmark-2.2/_index.md
```diff
@@ -588,8 +588,8 @@ grep -A 1 providers: /etc/kubernetes/encryption.yaml | grep aescbc
 
 The `EventRateLimit` plugin requires setting the `--admission-control-config-file` option and configuring details in the following files:
 
-- `/etc/kubernetes/admission.yaml`
-- `/etc/kubernetes/event.yaml`
+- `/opt/kubernetes/admission.yaml`
```
cloudnautique: Is this being done on Ubuntu or Rancher OS?

drpebcak: I did it on ubuntu. It might be worth noting that the location of these files may vary depending on the distro.

content/rancher/v2.x/en/security/benchmark-2.2/_index.md
```diff
@@ -0,0 +1,1794 @@
...
+**Result:** Pass
+
+#### 1.1.21 - Ensure that the `--kubelet-certificate-authority` argument is set as appropriate (Scored)
+
+**Notes**
+
+RKE is using the kubelet's ability to automatically create self-signed certs. No CA cert is saved to verify the communication between `kube-apiserver` and `kubelet`.
+
+**Mitigation**
+@TODO: See what happens when you give RKE a private cert.
```
cloudnautique: Is this in progress still?

drpebcak: Nope. We don't have an option for the kubelet's CA when using custom certs either.

catherineluse:
Hi, I was asked to review this PR so I just have a few questions/feedback:

Questions on hardening guide
- Is there a reason why the hardening guide starts with section 1.1 instead of 1? It seems like it's adding an extra level of hierarchy
- The beginning of the hardening guide talks about profile definitions. What is a profile for in this context?
- The hardening section has a link to the benchmark section. Should the benchmark section also have a link to the hardening section?

Questions on benchmark guide
- The “return value” header has nothing under it for some items in one section. Is there supposed to be a return value? This is in section 1.1.39 - "Ensure that the API Server only makes use of strong cryptographic ciphers (Not Scored)"
- In our benchmark guide, should we link to the official CIS benchmark guide or the website where people would download it? https://www.cisecurity.org/benchmark/kubernetes/
- I noticed that the hardening section talks about level 1 and 2 and the relative importance of each category of security measures. Should the benchmark section also explain those different levels?
- On the benchmark guide, it’s not always clear what the audit is for, or how to fix the code if it doesn’t pass the test. Can we explicitly say, “For the complete description of each audit, including rationales and remediation for failing tests, refer to the corresponding section of CIS_Kubernetes_Benchmark_v1.4.0.pdf”?

Proofreading stuff (really small changes that I could make)
- Fix Section 3.1.1 of the hardening guide where it refers to an image that isn’t showing up
- Spell out PSP on first usage in sections where it appears
- Spell out CIS the first time it is used on each section

I also found some places that need more backticks so I could add those as well.

drpebcak:
Feel free to make any changes you want. These are based off of the previous formatting for the 1.3 benchmark and the 2.1.x hardening guide.

> Which document would readers probably read first, the benchmark guide or the hardening guide?

Probably most people would just read the hardening guide, unless they were specifically interesting in the results of our benchmark self-assessment.

> Is there a reason why the hardening guide starts with section 1.1 instead of 1? It seems like it's adding an extra level of hierarchy

The benchmark uses the same concept. 1.1 is all `Rancher HA Kubernetes cluster host configuration` and then the third number (1.1.x) is the specific control that can be implemented.

> The beginning of the hardening guide talks about profile definitions. What is a profile for in this context?

It's a concept from the CIS benchmarks. A profile is a set of configurations that provide a certain amount of hardening (generally the more hardened an environment is, the more that affects performance).

> The hardening section has a link to the benchmark section. Should the benchmark section also have a link to the hardening section?

Probably not because the benchmark is purely supposed to address the CIS benchmark document. It isn't intended to give instructions on how to actually secure anything. The benchmark is just intended to show how an RKE Kubernetes cluster stacks up against the CIS benchmark.


> The “return value” header has nothing under it for some items in one section. Is there supposed to be a return value? This is in section 1.1.39 - "Ensure that the API Server only makes use of strong cryptographic ciphers (Not Scored)"
In our benchmark guide, should we link to the official CIS benchmark guide or the website where people would download it? cisecurity.org/benchmark/kubernetes
I noticed that the hardening section talks about level 1 and 2 and the relative importance of each category of security measures. Should the benchmark section also explain those different levels?
On the benchmark guide, it’s not always clear what the audit is for, or how to fix the code if it doesn’t pass the test. Can we explicitly say, “For the complete description of each audit, including rationales and remediation for failing tests, refer to the corresponding section of CIS_Kubernetes_Benchmark_v1.4.0.pdf”?

catherineluse:
The hardening guide mentions an image called no-local-cluster.png. I was trying to get it to show up on my local copy of the docs but I can't find the actual file, do you have it?

catherineluse:
Actually we can remove the reference to the image because there's an upcoming issue about removing UI screenshots going forward

drpebcak:
I don't have that image, looks like its broken on the live site for the 2.1 hardening guide as well.

Issue #1550, opened 2019-07-02
Document release branches of system-charts in Air-Gapped Instructions to "Configure Rancher System Charts"


axeal:
Document the branch of system-charts specific to each minor release at https://rancher.com/docs/rancher/v2.x/en/installation/air-gap-single-node/config-rancher-system-charts/

deniseschannon:
v2.0.0 - v2.2.3 | `master`
v2.2.4+ | `release-v2.2`
v2.3-alpha3 - `v2.3-preview1`
v2.3-alpha5 - `v2.3-preview2`

PR #1558, opened 2019-07-08
Specify system-charts branch for Rancher version in air gap docs

catherineluse:
This PR corresponds to this issue https://github.com/rancher/docs/issues/1550


deniseschannon:
@catherineluse As discussed, I've updated the release notes of all the releases so we should update the docs to ask them to refer to the release notes.

In the release notes, it's located within the "versions" section

Screen Shot 2019-07-15 at 1 50 01 PM

PR #1564, opened 2019-07-09
Add documentation about using IAM roles for S3 backup

moelsayed:
https://github.com/rancher/rancher/issues/21314


catherineluse:
The content looks good to me. I would want to fix the bullet formatting so that it looks right when the site is built, and add a link to AWS docs about how to give permission to a node.

deniseschannon:
@catherineluse this is merged into staging, so make a PR to staging and when its' ready, we can merge it into master.

PR #1577, opened 2019-07-15
Update azureAD install docs

dramich:
Azure has added a v2 api and changed the names of the fields, update the docs to reflect that.


dramich:
https://github.com/rancher/rancher/issues/21522

Issue #1578, opened 2019-07-15
Document that coredns is available as an option for k8s 1.12+ versions only and default and GA


deniseschannon:
Also remove any references of `experimental` in the docs.

Review docs to see where kubedns is used to show default. As of v2.2.5, coredns is default.

deniseschannon:
so coredns is the default provider in 1.14+

and optional for k8s 1.12, 1.13

PR #1581, opened 2019-07-16
Add CoreDNS as default and remove experimental

superseb:
https://github.com/rancher/docs/issues/1578


deniseschannon:
LGTM but do you think we need to add somewhere in the Rancher docs which DNS is default within each version?

superseb:
Added in https://github.com/rancher/docs/pull/1581/files#diff-3783c7f00db9dedcc8b78492fd669097R112

PR #1591, opened 2019-07-17
Edits to security hardening and benchmark guides

catherineluse:
This PR is a follow-up to https://github.com/rancher/docs/pull/1556

In this PR, formatting and style edits are the same in both versions of both the benchmark and hardening guides. In addition to that, the main security page lists both versions of each document.


catherineluse:
Closing this pr because it seems to mirror the same incorrect PR as last time instead of showing the new changes. Making a new branch instead

PR #1599, opened 2019-07-22
`repair-malformed-updates` config was completely removed in 1.14, so …

drpebcak:
…it should not be present in these guides.


drpebcak:
I don't have the ability to add reviewers here, so I'm going to mention everyone who was involved with the previous PR: @cloudnautique, @catherineluse, @deniseschannon, @oskapt.

PR #1602, opened 2019-07-23
Update example yaml for flannel network plug-in

catherineluse:
This PR corresponds to https://github.com/rancher/rancher/issues/16903

@loganhz and @galal-hussein , I have a question. Do the example yamls for canal and weave plugins have to be updated similarly to this update for the flannel plugin?


loganhz:
I'm not sure about it. We need input from @galal-hussein

galal-hussein:
@catherineluse @loganhz we don't really use `flannel_network_provider` in the cli, however they are being set only to be used from Rancher and exposed as UI options, instead we are using `plugin` and `option` in the cli

catherineluse:
Closing this PR because no doc change is needed https://github.com/rancher/rancher/issues/16903#issuecomment-526272384

PR #1616, opened 2019-07-25
Add a little extra delay between restarts with systemd

cryptk:
Documentation change to pair with rancher/k3s/pull/662


erikwilson:
LGTM, thanks!

PR #1621, opened 2019-07-26
fix incorrect s3backupconfig key

oskapt:
Sebastiaan helped me figure out that the Rancher docs have an incorrect key (`s3_backup_config`), whereas the RKE docs have it correct (`s3backupconfig`). There's no logging that says the key is invalid; it just ignored it. This fixes the mistake in the docs prior to releasing https://www.youtube.com/watch?v=PtYeoUT5gGk that references it.


deniseschannon:
duplicate https://github.com/rancher/docs/pull/1540/files

Issue #1478, opened 2019-06-05
Docs for feature: Cluster Templates/Policies


catherineluse:

catherineluse:
Closing this issue because https://github.com/rancher/docs/pull/1660 was merged

PR #1622, opened 2019-07-26
Cluster Templates docs

prachidamle:
https://github.com/rancher/docs/issues/1478

### DIFF
--- a/content/rancher/v2.x/en/user-settings/_index.md
+++ b/content/rancher/v2.x/en/user-settings/_index.md
@@ -13,6 +13,7 @@ The available user settings are:
 
 - [API & Keys]({{< baseurl >}}/rancher/v2.x/en/user-settings/api-keys/): If you want to interact with Rancher programmatically, you need an API key. Follow the directions in this section to obtain a key.gferfgre
 - [Cloud Credentials]({{< baseurl >}}/rancher/v2.x/en/user-settings/cloud-credentials/): Manage cloud credentials [used by node templates]({{< baseurl >}}/rancher/v2.x/en/cluster-provisioning/rke-clusters/node-pools/#node-templates) to [provision nodes for clusters]({{< baseurl >}}/rancher/v2.x/en/cluster-provisioning/rke-clusters). Note: Available as of v2.2.0. 
+- [Cluster Templates]({{< baseurl >}}/rancher/v2.x/en/user-settings/cluster-templates): Manage cluster templates [can be used to provision clusters]({{< baseurl >}}/rancher/v2.x/en/cluster-provisioning/rke-clusters).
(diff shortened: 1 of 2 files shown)


prachidamle:
@catherineluse this is the PR for Cluster Templates feature

alena1108:
We should also talk about:

* Questions/answers in more detail. Also I think putting more details on kubernetes versions would be good
* Templates enforcement
* RBAC. We should mention that by default templates can be created by the global admin only, and the way to give permissions to other users.

catherineluse:
I agree with Alena. Based on reading the design doc, I had the following feedback:

- We should explain how revisions work, saying that after you create a cluster, you can't change the template or update any of the settings defined in the template. You can only update it to a new revision of the same template
- We should say how to change the default template
- clustertemplate should be two words 
- We should include a list of the things you can define in a cluster template
- Should describe out-of-the-box templates
- We should compare and contrast the privileges of a global admin vs. a cluster admin vs. a regular user

Potential warnings that we might need to add:
- Might need a warning that says if you provision a cluster with the template, you will not be able to change anything that was defined by the template
- Might need a warning that says if global admin turned on template enforcement, you can’t provision a cluster without a template
- Might need a warning that says after you provision a cluster with a template, you can’t change the template. You can only update it to a new revision of the same template

deniseschannon:
@prachidamle Can you rebase so that this PR can go directly to `staging`? I've changed the base branch on the PR. 

can you address @alena1108's comments. @catherineluse can update and her her feedback after you've given the first draft on alena's comments.

prachidamle:
@deniseschannon I have rebased with staging, but the build is failing - not sure what could be a reason. @catherineluse  this is an updated draft. 
W.R.T enforcement, user roles, there are some changes to the design on cluster templates that I will update on the design docs. Also we are not including the out-of-the-box templates in this release.

deniseschannon:
@prachidamle i should have fixed the drone build (hopefully) by rebasing staging off of master.

catherineluse:
These changes look really good to me.

Based on the feature demo this morning, I have a couple of questions:

- Should we say that you can only create a cluster from a template if the cluster is created with RKE?
- Should we say that when you pick the Kubernetes version, if it ends in x, you are allowing the cluster to have any Kubernetes version within that major version?

prachidamle:
@catherineluse 
- yes I will add more on the .x k8sversion
- w.r.t mentioning about rke clusters - I think I already say it in the introductory line 

"When you provision a cluster [hosted by an infrastructure provider]({{< baseurl >}}/rancher/v2.x/en/cluster-provisioning/rke-clusters), cluster templates can be used to provision the cluster. "

Should we elaborate more on this?

deniseschannon:
@catherineluse we should clarify that this applies to "Rancher launched Kubernetes clusters", which is RKE clusters. maybe link to those type of clusters

We should clarify that for kubernetes versions that ends in `.x`, it will allow users to launch Rancher with the latest k8s maintenance version of that minor version. When editing the cluster, users will be able to upgrade to the latest available kubernetes maintenance versions. we might want to give a more explicit example

PR #1623, opened 2019-07-29
Update after docker-compose-agent.yml removal

yamt:
Also, stop repeating almost same examples twice.


erikwilson:
LGTM, thanks!

PR #1624, opened 2019-07-29
changing alpine to busybox in instructions

leodotcloud:
Addresses: https://github.com/rancher/rancher/issues/21650

Since we are including busybox image for fixing https://github.com/rancher/rancher/issues/20780, why not use the same for other purposes as well?


superseb:
I dont think there was a specific need for alpine, it was just picked randomly. For audit log, we use busybox, for rke-tools we use alpine. 

In this case, both images provide a tar binary that can do gzip.

```
$ docker run alpine tar 2>&1|head -1
BusyBox v1.30.1 (2019-06-12 17:51:55 UTC) multi-call binary.
$ docker run busybox tar 2>&1|head -1
BusyBox v1.31.0 (2019-07-16 01:13:11 UTC) multi-call binary.
```

PR #1639, opened 2019-08-04
Add tips for  Additional Trusted CAs process

geeckmc:


geeckmc:
I am using v2.2.6 and this error still exists in this version

geeckmc:
> This was fixed in v2.2.2, see [rancher/rancher#19473](https://github.com/rancher/rancher/issues/19473). What version are you using?

I am using v2.2.6 and this error still exists in this version

superseb:
Then it should be created as an issue and not as a workaround in the docs, can you check if the fix in the mentioned issue is applied and if you still experience issues, file an issue with the exact steps and commands used to reproduce?

Issue #1645, opened 2019-08-06
Rearrange HA installation docs so that system chart options are more visible


catherineluse:
Currently, there are some resources linked under the Install Rancher section for HA installation (https://rancher.com/docs/rancher/v2.x/en/installation/ha/helm-rancher/) that are also relevant if you want an air gap HA installation. However, the way it is now, users who are reading the air gap HA install would have difficulty finding these docs:

- Adding TLS Secrets
- Chart Options
- Troubleshooting

These docs should be moved to the base directory for the installation docs. Then the "install rancher" step for each tutorial can individually refer back to these resources.

catherineluse:
The required tools for air gap and regular HA installation are duplicated. They should probably be deduplicated as part of this reorganization.

catherineluse:
Closing this issue because the installation docs were rearranged

PR #1653, opened 2019-08-08
Move HA installation reference docs so air gap readers can see them

catherineluse:
This PR addresses this issue: https://github.com/rancher/docs/issues/1645


deniseschannon:
@catherineluse I'm going to close this PR as the changes in the install guide and v2.3 changes will make this difficult. I'll move the corresponding issue back into need to be worked on.

PR #1668, opened 2019-08-15
[k3s] Should download charts first when building from source

warmchang:
[k3s] Should download charts first when building from source (local build), otherwise the chartsTraefik tgz will not be compiled into the k3s binary.

xref: https://github.com/rancher/k3s/pull/729#issuecomment-521501318


erikwilson:
lgtm

PR #1676, opened 2019-08-18
Add "proxy_buffering off" to nginx reverse proxy config

dankeder:
Add "proxy_buffering off" to the nginx reverse proxy config example.

By default, proxy buffering is turned on and because of that nginx doesn't send the response until it decides to flush proxy buffers. That causes commands that wait for a response (such as `kubectl logs -f` or `kubectl get pods -w`) to not receive anything and just hang.


dankeder:
This is probably related to this issue: https://github.com/rancher/rancher/issues/14605

superseb:
That page is not kept up-to-date as its only supported up to v2.0.8, the example NGINX config can be found at https://rancher.com/docs/rancher/v2.x/en/installation/ha/helm-rancher/chart-options/#example-nginx-config. We should probably just link there to avoid duplicate data which can get out of sync.

dankeder:
All right. Google search for "rancher nginx reverse proxy" leads to this page. Btw. the page you link already contains `proxy_buffering off`. If I found it first it would have saved me a couple of hours troubleshooting :smile:.

I'll keep this PR open and you decide what happens to it. Either close it or merge it so it's synced with the new page. Thank you.

superseb:
I created the link in https://github.com/rancher/docs/pull/1682, thanks for reporting

Issue #1651, opened 2019-08-07
Update the etcd restoration process in RKE docs


galal-hussein:
The etcd snapshot restore was optimized in 0.2.x, to get rid off the extra steps for the restoration, the rke restoration currently does the following:
```
1- rke remove to clean old data
2- sync the snapshot (or download from s3)
3- restores the snapshot and starts etcd
4- triggers rke up
```
The docs need to explain the difference in details.

catherineluse:
Hi Galal, I have a couple of questions, just to clarify - 

1. Which steps are no longer necessary? Cleaning the nodes?
2. Which commands are required in 0.2.x?

PR #1679, opened 2019-08-19
Update the etcd restoration process in RKE docs

catherineluse:
This PR addresses the issue https://github.com/rancher/docs/issues/1651

The main reason for this PR is to clarify which steps are automated in RKE v0.2.0 that used to be a manual process in RKE prior to v0.2.0.

As I read through the existing RKE snapshot/restore document, I thought that as we elaborate on the differences between RKE versions, the document is going to get harder to read because everyone has to read about both versions of RKE. I also noticed that the example scenario part leaves out the `rke remove` command, kind of assuming RKE v0.2.0+ is being used (among other minor inconsistencies). 

Therefore, in an attempt to make the RKE restore docs easier to read, this PR includes the following changes:

- Breaks the RKE backup and restore document into sections on separate pages 

Screen Shot 2019-08-19 at 2 42 19 PM

- Divides content for different RKE versions into different tabs where the content is version-specific, making it clear at a glance that the older RKE version has more steps
Screen Shot 2019-08-19 at 2 35 15 PM
Screen Shot 2019-08-19 at 2 35 36 PM

- Updates the example scenario, so it's clear that the older RKE version has more steps
Screen Shot 2019-08-19 at 2 35 48 PM
Screen Shot 2019-08-19 at 2 36 07 PM


content/rke/latest/en/etcd-snapshots/example-scenarios/_index.md
```diff
@@ -0,0 +1,256 @@
...
+
+```
+root@node2:~# poweroff
+```
+
+|  Name |    IP    |          Role          |
+|:-----:|:--------:|:----------------------:|
+| node1 | 10.0.0.1 | [controlplane, worker] |
+| ~~node2~~ | ~~10.0.0.2~~ | ~~[etcd]~~                 |
+| node3 | 10.0.0.3 | [etcd]                 |
```
galal-hussein: can we remove node3 from the table since it will be added in the next section

catherineluse: Removed

Issue #1702, opened 2019-08-30
Hardening Guide - Rancher v2.2.x erroneously asks us to create files in /etc/kubernetes


Tejeev:
Every occurrence that instructs user to create a new file in /etc/kubernetes and any referencing these files should be corrected to point to /opt/kubernetes.

I believe it also needs to direct user to bind mount that in the various RKE containers via `extra_binds: `

If you create files in /etc/kubernetes, RKE will erase them whenever it installs or recovers a cluster.

Issue #1669, opened 2019-08-15
Fix hardening guide for nodes running RancherOS


janeczku:
https://rancher.com/docs/rancher/v2.x/en/security/hardening-2.2

Requirement:
There are multiple locations in the hardening guide where users are instructed to create files 
in the `/etc/kubernetes`directory, e.g.:

> touch /etc/kubernetes/admission.yaml

source: https://rancher.com/docs/rancher/v2.x/en/security/hardening-2.2/#1-1-4-place-kubernetes-event-limit-configuration-on-each-control-plane-host

When performing these steps on nodes running RancherOS, cluster creation will fail. RKE uses a different path to store Kubernetes configs on RancherOS: `/opt/rke/etc/kubernetes`.

Fix:
Call this out in the hardening guide wherever needed. E.g. users should do instead:

> > touch /opt/rke/etc/kubernetes/admission.yaml

PR #1704, opened 2019-08-30
Fix hardening guide instructions for k8s config files

drpebcak:
- Fixes Path (RKE will remove files from `/etc/kubernetes`)
- Mentions `extra_binds` (required to actually load in these files after they are written to the host)

Should fix https://github.com/rancher/docs/issues/1669, although this is not specific to RancherOS.
Also fixes https://github.com/rancher/docs/issues/1702


drpebcak:
@deniseschannon @cloudnautique Please review/merge this when you can.

Issue #1659, opened 2019-08-09
Drop unsupported cluster types from cluster cloning docs


dramich:
The link points to a section on how to modify hosted cluster exported configs which are not supported at this time. https://rancher.com/docs/rancher/v2.x/en/cluster-admin/cloning-clusters/#2-modify-cluster-config

catherineluse:
Closing because the issue is addressed by https://github.com/rancher/docs/pull/1718

PR #1718, opened 2019-09-04
Say that only clusters hosted by infrastructure providers can be cloned

catherineluse:
This PR addresses this issue: https://github.com/rancher/docs/issues/1659


dramich:
lgtm

kreempuff:
I don't think this ever got deployed. I still see it [here](https://rancher.com/docs/rancher/v2.6/en/cluster-admin/cloning-clusters/)

CleanShot 2021-11-29 at 11 23 28@2x

Issue #1655, opened 2019-08-08
Update Rancher docs for new Resources menu


tfiduccia:
In : 2.3

The project level main menu has changed in 2.3.  We added Resources menu that contains:
Workloads
HPA
Pipeline
Secrets
Config

![image](https://user-images.githubusercontent.com/11514927/62726220-94f38e00-b9cb-11e9-9782-8bd58fff7bb4.png)

Workloads now only contain: Workloads, Load Balancing, Service Discovery & Volumes
![image](https://user-images.githubusercontent.com/11514927/62726240-a76dc780-b9cb-11e9-9fce-84c1edb2ca4f.png)

Secrets now contain: Secrets, Certificates, Registry Credentials
![image](https://user-images.githubusercontent.com/11514927/62726267-bc4a5b00-b9cb-11e9-91b9-4ba48f31fb53.png)

deniseschannon:
`From the main navigation bar, choose **Resources**. In versions prior to v2.3.0, choose **Workloads** on the main navigation bar.`

PR #1812, opened 2019-10-04
Adding a registry has been moved to Resources > Secrets

b0r1sp:


superseb:
Thanks, this will be covered in https://github.com/rancher/docs/issues/1655 as we can't just change things as we have other versions that have the old structure still.

PR #1814, opened 2019-10-06
Adding helm init

dnoland1:
Helm commands will not work if helm init has not been run first. For example:
$ helm repo add jetstack https://charts.jetstack.io
Error: Couldn't load repositories file (/home/rancher/.helm/repository/repositories.yaml).
You might need to run `helm init` (or `helm init --client-only` if tiller is already installed)

Also, it doesn't hurt to run "helm init" a second time if user has run it in the past.


deniseschannon:
@dnoland1 we're about to release new air gap docs with the v2.3 release and it changes this whole section. can you review the staging branch on where this should be updated?

Issue #1733, opened 2019-09-06
Document etcd snapshot setting that makes file name windows-friendly 


alena1108:
We have to document the optional flag that got introduced with the fix for https://github.com/rancher/rancher/issues/22019

PR #1819, opened 2019-10-07
Document safe timestamps

moelsayed:
https://github.com/rancher/docs/issues/1733


deniseschannon:
@moelsayed I'm merging this to staging, but we're still missing the docs for the RKE CLI and ideally an example in the RKE CLI that would show this option. 

We'd need to include it in the snapshot/recovery pages: 
https://rancher.com/docs/rke/latest/en/etcd-snapshots/

deniseschannon:
Spoke with @moelsayed offline. Since RKE uses a different format for naming recurring snapshots, there is no need for a safe timestamp using the RKE CLI.

PR #1828, opened 2019-10-08
Update required Windows Server version

catherineluse:
In this PR I have updated the minimum required Windows Server version from 1903 to 1809. 

There is an outstanding issue in that we should also update the recommended AMI for AWS. In the hardware section it currently says:

> >- If you are using AWS, Rancher recommends *Microsoft Windows Server 2019 Base with Containers* as the Amazon Machine Image (AMI).
>- If you are using GCE, Rancher recommends *Windows Server 2019 Datacenter for Containers* as the OS image.


deniseschannon:
This needs to be rebased but otherwise LGTM

content/rancher/v2.x/en/cluster-provisioning/rke-clusters/windows-clusters/_index.md
```diff
@@ -84,15 +88,15 @@ We recommend the minimum three-node architecture listed in the table below, but
...
-Node 2  | Linux (Ubuntu Server 18.04 recommended)           | [Worker]({{< baseurl >}}/rancher/v2.x/en/cluster-provisioning/#worker-nodes) | Support the Rancher Cluster agent, Metrics server, DNS, and Ingress for the cluster
-Node 3  | Windows (Windows Server 2019 required)            | [Worker]({{< baseurl >}}/rancher/v2.x/en/cluster-provisioning/#worker-nodes) | Run your Windows containers
+| Node   | Operating System                                    | Kubernetes Cluster Role(s)                                                                                                                                                                                                                         | Purpose                                                                             |
+| ------ | --------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------- |
```
deniseschannon: No need to fix it, but when building a table, you only need 3 dashes to distinguish the header vs. bottom

So all you need is something like this, regardless of how long the content is. 
```
|---|---|---|---
```

PR #1846, opened 2019-10-11
Added note in quickstart, following issue https://github.com/rancher/k3s/issues/873

drannenberg:
Note added concerning k3s quick-start install following issue https://github.com/rancher/k3s/issues/873


davidnuzik:
We very much appreciate the input and PR; however, at this time we are looking into a better way to handle port requirements and offer suggestions in the docs in a different area. We plan to address this in the near future as we work to improve the docs.

Issue #1832, opened 2019-10-09
Document secrets encryption flow for RKE clusters


alena1108:
We have to document on how to turn on/off encryption for rke clusters, including the

* ability to pass custom config
* specifics on what happens on the backend when the encryption is turned on (secrets get auto rotated)

catherineluse:
Closing this issue because https://github.com/rancher/docs/pull/1884 was merged

PR #1884, opened 2019-10-15
Add rke documentaion for secrets encryption

moelsayed:
https://github.com/rancher/docs/issues/1832

### DIFF
--- a/content/rke/latest/en/config-options/secrets-encryption/_index.md
+++ b/content/rke/latest/en/config-options/secrets-encryption/_index.md
@@ -18,7 +18,7 @@ RKE provides users with two paths of configuration to enable at-rest data encryp
 
 Both configuration options can be added during initial cluster provisioning or by updating an exsiting cluster.
 
-To utilize this feature, a new is added to [Kube API service Configuration]({{< baseurl >}}//rke/latest/en/config-options/services/#kubernetes-api-server). A full custom configuration looks like this:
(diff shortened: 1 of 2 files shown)


alena1108:
@catherineluse applied some minor corrections to the PR, and forwarding it to you for further editing.

One important point - this document was created for rke cli originally. But the same kind of functionality can be leveraged via Rancher UI for Rancher provisioned clusters (RKE). 

* All the modifications need to be done via editing cluster.yml 
* `Key Rotation` section has to be reworked for Rancher part https://github.com/rancher/docs/pull/1884/files#diff-93aeb367b99ac54df5b1890839cd9109R81. There is no 'rotate' action on the cluster in Rancher UI/API, so the keys can be rotated using disable/enable `secrets_encryption_config`https://github.com/rancher/docs/pull/1884/files#diff-93aeb367b99ac54df5b1890839cd9109R105


cc @deniseschannon

catherineluse:
@alena1108 and @moelsayed , I have copy edited this section. In the last section, which is about how changing the encryption could make it hard to restore, I changed the section title to "How to Prevent Restore Failures after Rotating Keys." I also tried to add some paragraphs to spell out explicitly what you should do to prevent losing data. These are the additions:

> Therefore, we recommend that when you enable or disable encryption, or when you rotate keys, you should [create a snapshot]({{<baseurl>}}/rke/latest/en/etcd-snapshots/one-time-snapshots/) so that your backup requires the same keys that you have access to.

> This also means you should not rotate the keys during the restore process, because you would lose the encryption keys in `cluster.rkestate`.

Are these additions correct/ok?

catherineluse:
@deniseschannon Alena reviewed these last changes. I believe this PR can be merged now.

alena1108:
LGTM

PR #1892, opened 2019-10-16
Add default security policy

cloudnautique:


deniseschannon:
Also, this option isnt documented here: https://rancher.com/docs/rancher/v2.x/en/cluster-provisioning/rke-clusters/options/

Should it be?

catherineluse:
@SheilaghM I'm not sure whether this PR is still current/relevant, or if it needs more changes to be added.

PR #1894, opened 2019-10-16
make metadata subsections linkable

kinarashah:


catherineluse:
Hm, I tested a link to one of the subsections and it looks like it works, although it links to a place on the page that's just below the heading: https://rancher.com/docs/rancher/v2.x/en/admin-settings/k8s-metadata/#refresh-kubernetes-metadata

This PR can be merged, but I think you should create an issue on github.com/rancherlabs/website and say that the link icon should show up on level 1 headings.

Issue #1818, opened 2019-10-07
[Rancher FAQ] Poor formatting and may need updates


davidnuzik:
The Rancher FAQ parent page https://rancher.com/docs/rancher/v2.x/en/faq/ uses headings to separate out sections i.e. Cattle, Environments & Clusters, etc.

The problem is they are the same size headings and so it's hard to read. The questions should not be using such large headings (should be smaller or just bold text or something).

Additionally, is someone combing through FAQ's each release? We should review and see if this page needs any updates.

davidnuzik:
It looks like you can boldify headings i.e.:
```
# **Example bold heading**
```

So, perhaps that can be done to separate out the questions. This avoids having to reformat child pages as well that utilize the heading to ask questions.

davidnuzik:
The problem seems to be that there doesn't seem to be a diff between level 3 heading and level 4 heading. Might also be solved by just making sections larger heading like just one # for example instead of three.

deniseschannon:
@davidnuzik there are plans to do a design discussion of the docs site in general and i believe things like the fact that H3 and H4 are the same will be added as well.

catherineluse:
In this PR https://github.com/rancher/docs/pull/1901 the formatting was addressed. 

Some items were updated in that PR but I made another issue specifically for reviewing the all of the items: https://github.com/rancher/docs/issues/1913

PR #1901, opened 2019-10-17
Edit and update FAQ

catherineluse:
This PR addresses this issue: https://github.com/rancher/docs/issues/1818

On the main FAQ page, there were some questions that were outdated (they indicated that Windows and Istio were not officially supported) so I changed the answers to include links to the new feature documentation.

The main FAQ page also had a lot of questions that assume you are deciding whether to upgrade from Rancher v1.x to v2.x, so I moved them to a new page focused on that topic.

I have some outstanding questions about whether some items in the FAQ are correct. Are the following answers still true?

> **Will Rancher v2.x support Hashicorp's Vault for storing secrets?**
> Secrets management is on our roadmap but we haven't assigned it to a specific release yet. 

> **Are you planning on supporting Traefik for existing setups?**
> We don't currently plan on providing embedded Traefik support, but we're still exploring load-balancing approaches.

> **Can I import OpenShift Kubernetes clusters into v2.x?**
> Our goal is to run any upstream Kubernetes clusters. Therefore, Rancher v2.x should work with OpenShift, but we haven't tested it yet.

> **Are you going to integrate Longhorn?**
> Yes. Longhorn was on a bit of a hiatus while we were working on v2.0. We plan to re-engage on the project.

### DIFF
--- a/content/rancher/v2.x/en/faq/_index.md
+++ b/content/rancher/v2.x/en/faq/_index.md
@@ -51,7 +51,7 @@ At this time, we only support Docker.
 
 **Does Rancher v2.x support Calico, Contiv, Contrail, Flannel, Weave net, etc., for embedded and imported Kubernetes?**
 
-We provide the ability to use Calico, Canal, and Flannel, but always refer to the [Rancher Support Matrix](https://rancher.com/support-maintenance-terms/) on what is officially supported.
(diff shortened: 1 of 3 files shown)


deniseschannon:
@catherineluse let's make an issue that focuses on reviewing all the answers in the FAQ but otherwise I'll merge as is

Issue #1467, opened 2019-05-31
ingress-nginx updateStrategy


patrick0057:
ingress-nginx pods are set to only upgrade on delete.  Due to this setting you have to go in after upgrading your cluster and delete them to get on the correct version for your deployment.  This makes sense as you would want to make sure everything in your cluster upgrade is working before upgrading the ingress-nginx pods.  The issue is however that it isn't documented anywhere that this has to be done.

$ kubectl get daemonset nginx-ingress-controller -n ingress-nginx  -o yaml | grep -A1 updateStrategy
  updateStrategy:
    type: OnDelete

PR #1902, opened 2019-10-18
Say to check ingress-nginx upgrade policy

catherineluse:
I'm making this correction because it turns out the information in this issue is outdated. https://github.com/rancher/docs/issues/1467

Reference:
https://rancher.slack.com/archives/C6U4L2DS9/p1571242285016800

### DIFF
--- a/content/rancher/v2.x/en/cluster-admin/editing-clusters/_index.md
+++ b/content/rancher/v2.x/en/cluster-admin/editing-clusters/_index.md
@@ -83,7 +83,7 @@ As of Rancher v2.3.0, the Kubernetes metadata feature was added, which allows Ra
 
 For RKE v0.1.8+, the `ingress-nginx` YAML does not contain an upgrade policy, which makes the update strategy `RollingUpdate` by default. This policy allows the daemonset to restart the pods.
 
-For RKE prior to v0.1.8, the `updateStrategy` directive in the `ingress-nginx` YAML is set to `OnDelete.` In that case, you will need to delete these pods to get the correct version for your deployment.
+For RKE prior to v0.1.8, the `updateStrategy` directive in the `ingress-nginx` YAML was set to `OnDelete` because RKE did not yet support updating ingress and addons. In that case, you will need to delete these pods to get the correct version for your deployment.
 
 ### Adding a Pod Security Policy
 
--- a/content/rancher/v2.x/en/cluster-admin/editing-clusters/_index.md
+++ b/content/rancher/v2.x/en/cluster-admin/editing-clusters/_index.md
@@ -79,7 +79,7 @@ As of Rancher v2.3.0, the Kubernetes metadata feature was added, which allows Ra
 
 **Result:** Kubernetes begins upgrading for the cluster. During the upgrade, your cluster is unavailable.
 
-> **Note:** The `ingress-nginx` pods are set to only upgrade on delete. After upgrading your cluster, you need to delete these pods to get the correct version for your deployment. 
+> **Note:** Check the upgrade policy of the `ingress-nginx` pods. If the upgrade policy is set to **onDelete,** you will need to delete these pods to get the correct version for your deployment.
 
 ### Adding a Pod Security Policy
 
--- a/content/rancher/v2.x/en/cluster-admin/editing-clusters/_index.md
+++ b/content/rancher/v2.x/en/cluster-admin/editing-clusters/_index.md
@@ -79,7 +79,11 @@ As of Rancher v2.3.0, the Kubernetes metadata feature was added, which allows Ra
 
 **Result:** Kubernetes begins upgrading for the cluster. During the upgrade, your cluster is unavailable.
 
-> **Note:** Check the upgrade policy of the `ingress-nginx` pods. If the upgrade policy is set to **onDelete,** you will need to delete these pods to get the correct version for your deployment.
+### Upgrading ingress-nginx
+
+For RKE v0.1.8+, the `ingress-nginx` YAML does not contain an upgrade policy, which makes the update strategy `RollingUpdate` by default. This policy allows the daemonset to restart the pods.
+
+For RKE prior to v0.1.8, the `updateStrategy` directive in the `ingress-nginx` YAML is set to `OnDelete.` In that case, you will need to delete these pods to get the correct version for your deployment.
 
 ### Adding a Pod Security Policy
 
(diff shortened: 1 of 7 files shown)


Oats87:
I’m not sure this is adequate information to provide - we should have concrete conditions that would cause the on delete

catherineluse:
I saw in this comment https://github.com/rancher/rancher/issues/13146#issuecomment-386009823 that RKE is adding the `updateStrategy: OnDelete`. Still researching the reason.

Oats87:
I still don't believe we should blindly call this out -- there are versions of RKE + K8s floating around that don't explicitly set OnDelete. Perhaps we can call out a "From Version" or something like this.

catherineluse:
Hussein said prior to RKE v0.1.8, RKE did not support updating ingress or addons. Therefore it added the onDelete. RKE is able to update them in v0.1.8+.

galal-hussein:
@catherineluse @Oats87 Just to be clear, The current version of nginx ingress yaml doesn't contain update policy in its yaml which makes the upgrade policy by default rolling update, now pods can be upgraded since the rolling update policy allow the daemonset to restarting its pods

catherineluse:
OK I'm changing the note to a section in the same place:

### Upgrading ingress-nginx

> For RKE v0.1.8+, the `ingress-nginx` YAML does not contain an upgrade policy, which makes the update strategy `RollingUpdate` by default. This policy allows the daemonset to restart the pods.

> For RKE prior to v0.1.8, the `updateStrategy` directive in the `ingress-nginx` YAML is set to `OnDelete.` In that case, you will need to delete these pods to get the correct version for your deployment.

deniseschannon:
@galal-hussein @Oats87 can you feel revie2 again?

content/rancher/v2.x/en/cluster-admin/editing-clusters/_index.md
```diff
@@ -79,7 +79,11 @@ As of Rancher v2.3.0, the Kubernetes metadata feature was added, which allows Ra
 
 **Result:** Kubernetes begins upgrading for the cluster. During the upgrade, your cluster is unavailable.
 
-> **Note:** The `ingress-nginx` pods are set to only upgrade on delete. After upgrading your cluster, you need to delete these pods to get the correct version for your deployment. 
+### Upgrading ingress-nginx
+
+For RKE v0.1.8+, the `ingress-nginx` YAML does not contain an upgrade policy, which makes the update strategy `RollingUpdate` by default. This policy allows the daemonset to restart the pods.
+
+For RKE prior to v0.1.8, the `updateStrategy` directive in the `ingress-nginx` YAML is set to `OnDelete.` In that case, you will need to delete these pods to get the correct version for your deployment.
```
Oats87: I don't think that we ever explicitly set the updateStrategy, did we?

catherineluse:
This issue seems related https://github.com/rancher/docs/issues/2022

Should I add some info about this to the upgrade section too, as this user suggests?

content/rancher/v2.x/en/cluster-admin/editing-clusters/_index.md
catherineluse: Hussein said the updateStrategy was explicitly set in the older version of RKE because addons and ingress were not able to be updated through RKE.

But then in this PR https://github.com/rancher/rke/pull/680 , addon updates were supported and the updateStrategy was not explicitly set anymore.

catherineluse:
I have made some more changes to address this related issue https://github.com/rancher/docs/issues/2022

content/rancher/v2.x/en/cluster-admin/editing-clusters/_index.md
```diff
@@ -79,7 +79,11 @@ As of Rancher v2.3.0, the Kubernetes metadata feature was added, which allows Ra
 
 **Result:** Kubernetes begins upgrading for the cluster. During the upgrade, your cluster is unavailable.
 
-> **Note:** The `ingress-nginx` pods are set to only upgrade on delete. After upgrading your cluster, you need to delete these pods to get the correct version for your deployment. 
+### Upgrading ingress-nginx
+
+For RKE v0.1.8+, the `ingress-nginx` YAML does not contain an upgrade policy, which makes the update strategy `RollingUpdate` by default. This policy allows the daemonset to restart the pods.
+
+For RKE prior to v0.1.8, the `updateStrategy` directive in the `ingress-nginx` YAML was set to `OnDelete` because RKE did not yet support updating ingress and addons. In that case, you will need to delete these pods to get the correct version for your deployment.
```
Oats87: This is not actually true -- the updateStrategy was set to what was "default" by Kubernetes, so the qualification here is actually around RKE version + Kubernetes version. For example, 1.13 did OnDelete

Oats87: so this actually comes from the the API version, for a while it was `extensions/v1beta1` which defaulted to `OnDelete` and now it's `apps/v1` which defaults to `RollingUpdate` -- we need to find out when this switched

Oats87: https://github.com/rancher/kontainer-driver-metadata/commit/10592a4f69a2c3b30855e8e0cf5f4a6ceefeba96#diff-1144caa68cad35fbcc94b0e38242934b this is when this appears to have changed, so when RKE started supporting 1.16...

catherineluse: I have changed that first paragraph. Does this capture it in the right amount of detail?

> For RKE v0.1.8+, the `ingress-nginx` YAML does not contain an `updateStrategy`. Therefore, the `updateStrategy` is the default value that is set by the Kubernetes API version. For example, the default `updateStrategy` for Kubernetes v1.13 is `onDelete`. When RKE began supporting Kubernetes v1.16, the `updateStrategy` for `ingress-nginx` became `RollingUpdate` by default for RKE clusters. This policy allows the daemonset to restart the pods.

Oats87: @catherineluse I think it would actually simplify the problem if we called it out like:

>>>Clusters that were created before Kubernetes 1.16 will have a strategy of `OnDelete`. Clusters that were created with Kubernetes 1.16 or newer will have `RollingUpdate`

catherineluse: ok, here's the shortened section:

> ### Updating ingress-nginx

> Clusters that were created before Kubernetes 1.16 will have an `ingress-nginx` `updateStrategy` of `OnDelete.` Clusters that were created with Kubernetes 1.16 or newer will have `RollingUpdate`.

> If the `updateStrategy` of `ingress-nginx` is `OnDelete,` you will need to delete these pods to get the correct version for your deployment.

PR #1911, opened 2019-10-21
Update RKE upgrade documentation

axeal:
Call out use of kubernetes_version as preferred method of upgrading Kubernetes version vs system_images, add additional details on presence of cluster state file


superseb:
@axeal please update/respond

PR #1923, opened 2019-10-23
Added minor reference to installing RKE.  Might still get missed

Tejeev:


catherineluse:
It might make more sense to have the prerequisites section on that page instead of the parent page. https://github.com/rancher/docs/blob/8d53a2a48d780f4dab76fb3be80f12318ca7bd5f/content/rancher/v2.x/en/installation/air-gap/_index.md

Tejeev:
> It might make more sense to have the prerequisites section on that page instead of the parent page. https://github.com/rancher/docs/blob/8d53a2a48d780f4dab76fb3be80f12318ca7bd5f/content/rancher/v2.x/en/installation/air-gap/_index.md

I think that makes a lot of sense

PR #1929, opened 2019-10-23
Update _index.md

albal:
It seems in Rancher 2.3 Workloads in the main menu bar should be clicked.


deniseschannon:
@albal The main menu bar was changed to "Resources" from Workloads.

PR #1930, opened 2019-10-23
Fixed audit example for protect-kernel-defaults and make-iptables-util-chains

dnoland1:


drpebcak:
@dnoland1 can you fix this in the older versions of the hardening guide as well? Looks like the same mistake was made each time.

dnoland1:
@drpebcak I fixed 2.1 and 2.2 as well.

PR #1936, opened 2019-10-25
Add baremetal quickstart

mkinney:


catherineluse:
Can you please generalize this PR so that it just says to use any modern Linux distribution instead of RancherOS?

mkinney:
Doesn't "bare metal" imply that there really is no OS, such as linux?

catherineluse:
Yeah, but the PR includes Rancher OS in the prerequisites and says to install it.

mkinney:
The prereqs say to download the iso file.

Do you have any suggestions on what to change so the instructions are more understandable?

catherineluse:
The problem isn't a lack of clarity. The problem is that it says to install the RancherOS ISO. My concern is that a newcomer reading these instructions might assume that because it says to install RancherOS, that means RancherOS is required to install Rancher. That would convey a different message than what's in our installation requirements, which say the following:

> Rancher should work with any modern Linux distribution and any modern Docker version. Rancher has been tested and is supported with Ubuntu, CentOS, Oracle Linux, RancherOS, and RedHat Enterprise Linux.

In the big picture, we want to show that Rancher is compatible with a wide variety of other technologies and we don't want people to think you have to use Rancher with other Rancher products, such as RancherOS.

catherineluse:
@SheilaghM The outstanding question on this PR is whether or not bare metal installation instructions should be included to the docs.

If this PR is merged, it would need to be modified to remove all mention of any specific Linux distribution, including RancherOS. RancherOS is deprecated, and as for the other distros, we don't want to say which versions Rancher was tested on or supported on in the docs, because the terms of service needs to be the single source of truth for that information.

Issue #1835, opened 2019-10-10
Document enabling forward host headers for NGINX 0.25.0 in Rancher Launched Kubernetes clusters in v2.3.0 and higher


superseb:
See https://github.com/rancher/rancher/issues/23341#issuecomment-540601760

Document the required change in https://rancher.com/docs/rancher/v2.x/en/installation/ha/helm-rancher/chart-options/#external-tls-termination

alena1108:
@sangeethah can someone from qa validate and confirm the workaround steps? So we can follow up documenting them.

catherineluse:
Closing this issue because https://github.com/rancher/docs/pull/1939 was merged

PR #1939, opened 2019-10-25
Document enabling forward host headers for NGINX 0.25.0

catherineluse:
This PR addresses this issue https://github.com/rancher/docs/issues/1835

### DIFF
--- a/content/rancher/v2.x/en/installation/ha/helm-rancher/chart-options/_index.md
+++ b/content/rancher/v2.x/en/installation/ha/helm-rancher/chart-options/_index.md
@@ -156,9 +156,9 @@ You may terminate the SSL/TLS on a L7 load balancer external to the Rancher clus
 
 Your load balancer must support long lived websocket connections and will need to insert proxy headers so Rancher can route links correctly.
 
-#### Enabling Forward Host Headers for NGINX v0.25.0
+#### Configuring Ingress for External TLS when Using NGINX v0.25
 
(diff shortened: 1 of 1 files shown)


content/rancher/v2.x/en/installation/ha/helm-rancher/chart-options/_index.md
```diff
@@ -156,6 +156,22 @@ You may terminate the SSL/TLS on a L7 load balancer external to the Rancher clus
 
 Your load balancer must support long lived websocket connections and will need to insert proxy headers so Rancher can route links correctly.
 
+#### Enabling Forward Host Headers
+
+_For Rancher v2.3.0+, which uses NGINX 0.25.0_
```
deniseschannon: Just reference for anyone using nginx 0.25.0.

catherineluse: updated

deniseschannon:
@superseb Does this only apply to Rancher installs or would we need this in the RKE docs as well for anyone using the nginx ingress controller?

superseb:
The reason we add it here is because it applies to a setup of Rancher we describe and it doesn't work in this newer version. I dont think we used to describe all changes in components that are used when setting up a cluster (or basically any component version besides Kubernetes), and all components provide changelogs, maybe we should link those in the docs.

PR #1957, opened 2019-10-31
Say that Windows Server 1809 also works with Docker 19.03

catherineluse:
Sowmya had a question about this and I confirmed it with Logan. Docker 19.03 is fine to use with Windows Server 1809, so I have updated the Windows requirements to reflect this.


catherineluse:
> We also need to update this page: https://rancher.com/docs/rancher/v2.x/en/installation/requirements/

Updated the installation requirements page

PR #1958, opened 2019-10-31
Add pre-reqs section to node-requirements

davidnuzik:
This just adds a note about how hostnames must be unique.
I couldn't think of anything else vital at this time but we may add a note about iptables in the Known Issues in a separate PR.

Resolves https://github.com/rancher/k3s/issues/857


content/k3s/latest/en/installation/node-requirements/_index.md
```diff
@@ -7,6 +7,9 @@ k3s is very lightweight, but has some minimum requirements as outlined below.
 
 Whether you're configuring a k3s cluster to run in a single-node or high-availability (HA) setup, each node running k3s should meet the following minimum requirements. You may need more resources to fit your needs.
 
+## Pre-requisites
+*    Two nodes cannot have the same hostname. If all your nodes have the same hostname, pass `--node-name` or `$K3S_NODE_NAME` respectively with a unique name for each node you add to the cluster.
```
ShylajaDevadiga: Could we remove respectively here as we don't use both the options at one time

PR #1959, opened 2019-10-31
Remove load balancer step for ha install

davidnuzik:
- Fix a couple of minor things in the ha installation _index.md
- Remove the "Create a load balancer..." step
- Remove the "Create a load balancer..." section
- Modify the "Join Worker Nodes" section accordingly

### DIFF
--- a/content/k3s/latest/en/installation/ha/_index.md
+++ b/content/k3s/latest/en/installation/ha/_index.md
@@ -43,8 +43,8 @@ Following the [Node Requirements]({{< baseurl >}}/k3s/latest/en/installation/nod
 Run the following command to join a worker node to the master nodes. You can get the node-token from any of the servers at `/var/lib/rancher/k3s/server/node-token`
 
 ```
-curl -sfL https://get.k3s.io | K3S_URL=https:/<Master_node>:6443 K3S_TOKEN=XXX sh-
+curl -sfL https://get.k3s.io | K3S_URL=https:/<master_node>:6443 K3S_TOKEN=XXX sh-
 ```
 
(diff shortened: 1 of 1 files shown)


content/k3s/latest/en/installation/ha/_index.md
```diff
@@ -36,19 +35,16 @@ On the second machine, run the following command. Since we ran the first node wi
...
 
 Run the following command to join a worker node to the master nodes. You can get the node-token from any of the servers at `/var/lib/rancher/k3s/server/node-token`
 
 ```
-curl -sfL https://get.k3s.io | K3S_URL=https:/<FQDN_of_LB>:6443 K3S_TOKEN=XXX sh-
+curl -sfL https://get.k3s.io | K3S_URL=https:/<Master_node>:6443 K3S_TOKEN=XXX sh-
 ```
 
-Provide the FQDN of the Load Balancer for the master nodes.
+Provide the IP or DNS in place of `<Master_node>` this can be any one master node. k3s automatically handles load balancing the master nodes. The K3S_URL you provide will actually be the last used in a fail-over scenario.
```
ShylajaDevadiga: We may not need the line K3S_URL you provide will actually be the last used in a fail-over scenario. It may be confusing when we have no log of the command used at a later time.

ShylajaDevadiga: master node instead of  `<Master_node>`

davidnuzik: Agreed

davidnuzik: fixed both to be <master_node> (lower case) I prefer to keep the _ because it's used in a cli example and we normally do this elsewhere in docs.

Issue #1018, opened 2018-11-28
Self-signed Gitlab Integration for Pipeline and other settings in Pipeline setting pages


deniseschannon:

MBishop17:
ask @gitlawr about this. We have docs about GitLab, but not about self-signed certs for it.

Issue #1017, opened 2018-11-28
Catalog Integration in Pipeline


deniseschannon:

MBishop17:
Ask Lawrence about this.

PR #1976, opened 2019-11-07
Update k3s main page

davidnuzik:
Update the main page for #1017 and #1018
This tweaks the main page to update the description of what we remove and add to k3s that sets us apart from k8s.

Closes https://github.com/rancher/k3s/issues/1017 and closes https://github.com/rancher/k3s/issues/1018

### DIFF
--- a/content/k3s/latest/en/_index.md
+++ b/content/k3s/latest/en/_index.md
@@ -20,14 +20,12 @@ What is this?
 
 k3s is intended to be a fully compliant Kubernetes distribution with the following changes:
 
-1. Removed legacy and non-default features. You lilkely won't notice the
-   stuff that has been removed.
-2. Removed most in-tree plugins (cloud providers and storage plugins) which can be replaced
+1. Removed most in-tree plugins (cloud providers and storage plugins) which can be replaced
    with out of tree addons.
-3. Added sqlite3 as the default storage mechanism and support for other external SQL databases such as PostgreSQL and MySQL. etcd3 is also supported as an external database.
-4. Added local storage provider, service load balancer, helm-controller, and traefik ingress controller.
-5. Wrapped in simple launcher that handles a lot of the complexity of TLS and options.
-6. Minimal to no OS dependencies (just a sane kernel and cgroup mounts needed). k3s packages required dependencies
+2. Added sqlite3 as the default storage mechanism and support for other external SQL databases such as PostgreSQL and MySQL. etcd3 is also supported as an external database.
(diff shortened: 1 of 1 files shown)


content/k3s/latest/en/_index.md
```diff
@@ -20,14 +20,14 @@ What is this?
 
 k3s is intended to be a fully compliant Kubernetes distribution with the following changes:
 
-1. Legacy, alpha, non-default features are removed. Hopefully, you shouldn't notice the
+1. Removed legacy and non-default features. You lilkely won't notice the
```
davidnuzik: Per discussion with @cjellick  essentially the whole line here actually needs to be removed.

catherineluse: agree

catherineluse: But we still want to mention that legacy and non-default features are removed right? even though that's not all that changed?

content/k3s/latest/en/_index.md
```diff
@@ -20,14 +20,14 @@ What is this?
...
    stuff that has been removed.
 2. Removed most in-tree plugins (cloud providers and storage plugins) which can be replaced
    with out of tree addons.
-3. Add sqlite3 as the default storage mechanism. etcd3 is still available, but not the default.
-4. Wrapped in simple launcher that handles a lot of the complexity of TLS and options.
-5. Minimal to no OS dependencies (just a sane kernel and cgroup mounts needed). k3s packages required
-   dependencies
+3. Added sqlite3 as the default storage mechanism and support for other external SQL databases such as PostgreSQL and MySQL. etcd3 is also supported as an external database.
```
catherineluse: I would add to the database section a sentence about the changed architecture "Instead of using nodes with an etcd Kubernetes role, the server updates data in an external database, such as PostgreSQL, MySQL, or etcd3."

content/k3s/latest/en/_index.md
davidnuzik: No, actually legacy and non-default features are _not_ removed any more. So all three, Legacy, alpha, and non-default features are still in.
Thus, we do indeed want to remove this whole line. Will fix in incoming commit.

content/k3s/latest/en/_index.md
davidnuzik: Unless I'm mistaken I think based on our convo earlier this isn't needed. Please correct me if I'm wrong.

catherineluse: Yeah this is fine as is.

catherineluse:
LGTM

PR #2004, opened 2019-11-14
Typofix

soerface:


erikwilson:
Thanks @soerface! It looks like the fixed block of text has been removed, so closing this PR.

PR #2015, opened 2019-11-19
Support secure AWS Task IAM Roles

sodre:
The Task IAM Rules were extracted verbatim from the Amazon documentation [steps 5, 6, and 7]( https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-install.html)

The latter two rules were adapted from [Amazon documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/instance_IAM_role.html) so that system-docker traffic retains access to the instance-metadata, while traffic from user-docker does not.

This documentation change is not sufficient to get Task IAM Roles to work, it requires a modification to the amazon-ecs-agent service, which will follow in a separate PR to the `os-services` repo.


sodre:
Hi!

Any updates on getting this issue fixed?

catherineluse:
@SheilaghM This PR is dependent on a RancherOS change to be merged, but since RancherOS is deprecated, I'm wondering if this PR should be closed.

Issue #1993, opened 2019-11-12
Update Istio docs to include no sidecar annotation


cloudnautique:
On the enable istio sidecar injection in a namespace page: https://rancher.com/docs/rancher/v2.x/en/cluster-admin/tools/istio/setup/enable-istio-in-namespace/

We should include that if you need to exclude a workload from getting the sidecar injected the user can use the following annotation:
`sidecar.istio.io/inject: “false”`

catherineluse:
Closing this issue because it is addressed by https://github.com/rancher/docs/pull/2024

PR #2024, opened 2019-11-22
Explain how to exclude workloads from Istio sidecar

catherineluse:
This PR addresses this issue: https://github.com/rancher/docs/issues/1993


catherineluse:
@cloudnautique If you're doing it through the UI, do you need to include the quotation marks around the "false"?

PR #2025, opened 2019-11-22
Clarification of where to run

Tejeev:


catherineluse:
I've edited this PR to make the same change to the other tab for the older version of RKE, because it applies to that one too.

Looks good to me

PR #2028, opened 2019-11-22
Document how to tell Rancher to ignore a node

catherineluse:
I reorganized this page because I'm about to add a new section to it, and I don't want to just tack it on, because in the long term that could make it cluttered.

So I added a table of contents and tried to group similar items together, just to prepare it for taking a new section.

I made two corrections in the organization commit as well:

- There was a note warning people that "Currently, the timeout setting is not enforced while draining a node. This issue will be corrected as of Kubernetes 1.12." That version of k8s is in the past, so I changed it to "The timeout setting was not enforced while draining a node prior to Kubernetes 1.12."
- There were two misleading references to node pools, saying that they "automatically" replace deleted nodes. I clarified that to say that they will only automatically replace nodes if you enable node auto-replace.


catherineluse:
I'm trying to pin down some details but this is the draft of the new section:

# Labeling a Node to be Ignored by Rancher

_Available as of 2.3.3_

Some solutions, such as F5's BIG-IP integration, may require creating a node that is never registered to a cluster.

Since the node will never finish registering, it will always be shown as unhealthy in the Rancher UI.

In that case, you may want to label the node to be ignored by Rancher so that Rancher only shows nodes as unhealthy when they are actually failing.

You can label nodes to be ignored by using a setting in the Rancher UI, or by using `kubectl`.

### Labeling Nodes to be Ignored with the Rancher UI

To add a node that is ignored by Rancher,

1. From the **Global** view, click the **Settings** tab.
1. Go to the `ignore-node-name` setting and click **Ellipsis (...) > Edit.**
1. Enter a name that Rancher will use to ignore nodes. All nodes with this label will be ignored.
1. Click **Save.**

**Result:** Rancher will not wait to register nodes with this name. In the UI, the node will displayed with a grayed-out status.

If the setting is changed afterward, the ignored nodes will continue to be hidden.

### Labeling Nodes to be Ignored with kubectl

To add a node that will be ignored by Rancher, use `kubectl` to create a node that has the following label:

```
cattle.rancher.io/node-status: ignore
```

**Result:** Rancher will not attempt to sync with this node. 

If the label is added before the node is added to the cluster, the node will not be shown in the Rancher UI. 

If the label is added after the node is added to a Rancher cluster, the node will not be removed from the UI.

If you delete the node from the Rancher server using the Rancher UI or API, the node will not be removed from the user cluster if the `nodeName` is listed in the Rancher settings under `ignore-node-name`.

catherineluse:
This should be considered a draft PR

catherineluse:
# Viewing the changes in a web browser
```
git clone https://github.com/catherineluse/docs.git
cd docs
git fetch -a
git checkout ignore-node
./scripts/dev
```
Then view `http://localhost:9001/rancher/v2.x/en/cluster-admin/nodes`

PR #2072, opened 2019-12-10
add supported docker versions

ebauman:
Terms of service mention that we support both RHEL Docker as well as upstream Docker. However these installation docs did not have that listed.


davidnuzik:
- Made sure QA is indeed testing these versions in RHEL/CentOS
- Reviewed changes

LGTM

catherineluse:
Denise said we're going to entirely remove the table of supported OS/Docker versions from the installation docs, because maintaining the table in two places (installation docs and terms of services) has led to several discrepancies between them. This section will just say that Rancher should work with any modern version of Linux and Docker

catherineluse:
Closing this pull request because we're removing the specific OS and docker versions from this doc

PR #2073, opened 2019-12-10
Re-add Helm 2 docs on separate tab

catherineluse:
I've been asked to re-add Helm 2 docs, but this time it's on a separate tab

### DIFF
--- a/content/rancher/v2.x/en/installation/ha/helm-rancher/_index.md
+++ b/content/rancher/v2.x/en/installation/ha/helm-rancher/_index.md
@@ -9,9 +9,9 @@ For systems without direct internet access, see [Air Gap: High Availability Inst
 
 Refer to the [Helm version requirements]({{<baseurl>}}/rancher/v2.x/en/installation/helm-version) to choose a version of Helm to install Rancher.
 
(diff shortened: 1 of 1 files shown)


content/rancher/v2.x/en/installation/ha/helm-rancher/_index.md
```diff
@@ -85,13 +154,27 @@ These instructions are adapted from the [official cert-manager documentation](ht
...
-1. Install the cert-manager Helm chart
-    ```plain
-    helm install \
-      cert-manager jetstack/cert-manager \
-      --namespace cert-manager \
-      --version v0.9.1
-    ```
+1. Install the cert-manager Helm chart. The command is different for Helm 2 and Helm 3.
+
+{{% tabs %}}
```
deniseschannon: If we have accordions for the install tiller part, maybe this should be an accordion, too. 

Thoughts?

catherineluse: My reasoning for the accordions on Tiller steps is that there's no equivalent for Helm 3. But in the other installation commands, there is a Helm 3 equivalent which we can show on the default tab. This way, the Helm 2 steps are hidden by default in the accordion and non-default tabs

catherineluse:
Closing this PR because after some discussion, it sounds like this way of writing it is confusing. It would be better to have the Helm 2 docs completely separate.

PR #2089, opened 2019-12-12
Document audit log feature of RKE

catherineluse:


content/rke/latest/en/config-options/audit-log/_index.md
```diff
@@ -0,0 +1,132 @@
...
+  kube-api:
+    audit_log:
+      enabled: true
+      configuration:
+        max_age: 6
+        max_backup: 6
+        max_size: 110
+        path: /var/log/kube-audit/audit-log.json
+        format: json
+        policy:
```
leodotcloud: The policy defined here is just a copy of the example provided on k8s docs. It shows various options, one doesn't need to specify all of these to customize stuff.

catherineluse: I added a note explaining the source of the example

Issue #1304, opened 2019-03-22
Make it clear in docs that unusable namespaces are possible to be created in kubctl


jambajaar:
Refer 
https://github.com/rancher/rancher/issues/18615

Looks like there’s no way to prevent kubectl from creating unusable namespaces when using Rancher’s project member role.

Can this be made clear in documentation.

The above comment is based on Rancher v2.1.3

PR #2142, opened 2019-12-30
Say how to avoid creating unusable namespaces with kubectl

catherineluse:
This PR is intended to address this issue https://github.com/rancher/docs/issues/1304

@jambajaar and @janeczku , does this note convey the problem and how to avoid it?

> **Note:** If you create a namespace with `kubectl`, it may be unusable because `kubectl` doesn't require your new namespace to be scoped within a project that you have access to. If your permissions are restricted to the project level, it is better to [create a namespace through Rancher]({{<baseurl>}}/rancher/v2.x/en/project-admin/namespaces/#creating-namespaces) to ensure that you will have permission to access the namespace.


janeczku:
> Although you can assign role-based access to namespaces in the base version of Kubernetes, you cannot assign roles to namespaces in Rancher. Instead, assign role-based access at the project level.

Maybe something like:

> To manage permissions in a vanilla Kubernetes cluster, cluster admins configure role-based access policies for each namespace. With Rancher user permissions are assigned on the project level instead and automatically inherited by any namespace owned by the particular project.

Issue #1968, opened 2019-11-06
command to show available Kubernetes versions


nickngch:
https://rancher.com/docs/rke/latest/en/upgrades/

provide the command to list out the available Kubernetes versions

`rke config -l -a`

Useful when the user wants to specify the Kubernetes version in cluster.yaml file

superseb:
This should be in after the requested changes has been added in https://github.com/rancher/docs/pull/1911

catherineluse:
Closing this issue because the command was moved to the upgrades page in this PR https://github.com/rancher/docs/pull/2143

PR #2143, opened 2019-12-30
Add RKE command to show available K8s versions

catherineluse:
This PR is intended to address this issue:
https://github.com/rancher/docs/issues/1968

### DIFF
--- a/content/rke/latest/en/config-options/_index.md
+++ b/content/rke/latest/en/config-options/_index.md
@@ -60,22 +60,9 @@ ignore_docker_version: true
 
 ### Kubernetes Version
 
-By default, RKE is defaulted to launch with a specific Kubernetes version. You can also select a different version of Kubernetes to install for your cluster. Each version of RKE has a specific list of supported Kubernetes versions.
+For information on upgrading Kubernetes, refer to the [upgrade section.]({{<baseurl>}}/rke/latest/en/upgrades/)
 
-You can set the Kubernetes version as follows:
-
-```yaml
-kubernetes_version: "v1.11.6-rancher1-1"
-```
-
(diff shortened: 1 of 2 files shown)


catherineluse:
@superseb The link to the section about getting the supported Kubernetes versions was already on the RKE upgrades page. Are you saying that nothing needs to be added to that page?

superseb:
The PR that updated the Upgrading section contained the necessary information to upgrade, the question is if you want the command on a page that is linked on there, or if you want the command on the page it self (which caused the "issues" mentioned above).

And the other two things were not in your PR, just pointing out something that can be fixed straight away as I just saw that (and apparently not at the time I reviewed it previously)

catherineluse:
@superseb After thinking about it some more, I think the Kubernetes versions command seems out-of-place in the config options section because it isn't really a config option.

Therefore I have moved that section into the upgrades section and reorganized the upgrades section to make it a little bit cleaner. Please let me know if this looks OK.

superseb:
It does look clean this way, the only "problem" I see is that we now still describe setting the kubernetes version in `cluster.yml` under config options, but the command to help you with the value for that is now under upgrades and is linked there and that feels the other way around. The other use of a similar command is under `config-options/private-registries`.

The flow that I see now is that if a new user wants to configure his/her `cluster.yml` with `kubernetes_version`, it ends up at docs root -> Config Options -> Kubernetes Version -> Upgrades -> Listing Versions. In both cases, the information needed is available. I'm fine with the current approach as this is the best user experience for configuring this.

catherineluse:
I think that the Kubernetes upgrade information should only be included on the upgrades page, because downgrades are not supported.

I think the Config Options -> Kubernetes Version section should just say "For information on upgrading Kubernetes, refer to the upgrades section. Downgrading Kubernetes is not supported."

Issue #2153, opened 2019-12-31
Clarify requirements for Istio Visualizations access


rancher-max:
https://rancher.com/docs/rancher/v2.x/en/cluster-admin/tools/istio/rbac/#access-to-visualizations should specify that configuring permissions for Jaeger and Kiali requires users to also have (at least) Read Only access to System project

rancher-max:
This was my mistake and can be ignored!! Sorry!

PR #2161, opened 2020-01-03
Read-only system project access is needed for configuring Istio viz access

catherineluse:
This PR addresses this issue https://github.com/rancher/docs/issues/2153


rancher-max:
@catherineluse Can we revert this? This was my mistake 😦 It looked like it needed this in v2.3.3, but there was a different setting that I had enabled improperly that lead to this. I'm so sorry!!

catherineluse:
Ok, no problem. The note has been deleted https://github.com/rancher/docs/pull/2163

PR #2167, opened 2020-01-08
add comment for additional network plug-ins

bcreane:
Update network plug-in comment to include options in addition to "canal." Follows the same format as the following comment for DNS provider.


bcreane:
@catherineluse - thanks for the review + merge!

PR #2173, opened 2020-01-14
Clarified where possible

Tejeev:
Clarified and added a space, moved the result to the end of the instructions to match prior pages and intuitive format, and swapped 'cluster' for 'project' as there is no import yaml option in the cluster view.  This last one could do with some investigation as I'm not sure if it was maybe saying to go to the project, import the yaml, and choose the Cluster 'Import Mode' (Options are Cluster, Project, and Namespace).  I'm not sure which should be chosen and it might be worth clarifying on all YAML import steps.


catherineluse:
Yes, I believe it is correct to import from the project view and then use the default option from there, which is "Project: Import resources into this project."

PR #2195, opened 2020-01-21
EIO-247: CIS-1.5 benchmark self assessment

paraglade:


Rancheroo:
Look's great for the new page, however
What about the links that are already there below (see bottom of comment)? Will these be cleaned up? as they talk about using Hardening Guide for Rancher 2.3.3+ with Kubernetes 1.16 (CIS has no ratings for 1.16, that i can see) and the guide does not match the 1.15 CIS benchmark. So super confusing for customers.

https://rancher.com/docs/rancher/v2.x/en/security/hardening-2.3/
https://rancher.com/docs/rancher/v2.x/en/security/hardening-2.3.3/

PR #2309, opened 2020-02-26
Add Shibboleth first draft

catherineluse:


nikkelma:
Are the OpenLDAP sections intended to be included in this section?

catherineluse:
Luther and I changed the Shibboleth docs since this PR. We started using open LDAP docs as a base because we know that OpenLDAP can be the back end of Shibboleth and can allow you to use groups in Rancher. The new version is here https://staging.rancher.com/docs/rancher/v2.x/en/admin-settings/authentication/shibboleth/ and it might be easier to leave feedback in a new doc issue

nikkelma:
I skimmed through the docs on staging before commenting, but didn't notice the following:
> If you also configure OpenLDAP as the back end to Shibboleth

Thanks for the clarification @catherineluse, I'm all set!

PR #2311, opened 2020-02-28
Fix wrong apiVersion for PodSecurityPolicy

caiconkhicon:
To create a PodSecurityPolicy, the apiVersion `policy/v1beta1` should be used. If I apply the current manifest, which use `apiVersion: policy/v1beta1`,  it throws out the error: 
`no matches for kind "PodSecurityPolicy" in version "extensions/v1beta1`


paraglade:
This setting was deprecated since v1.10 of kubernetes and `apiVersion: policy/v1beta1` is the version that should be used on k8s 1.16 and greater.

https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

PR #2321, opened 2020-03-02
RKE: Added documentation to install it using MacPorts, and refactored MacOS into macOS

emcrisostomo:
I've updated the documentation to add MacPorts as a way to install `rke` on macOS. I've also taken advantage of this update to refactor `MacOS X` into `macOS` (in that page).

Cheers


catherineluse:
We don't support installing RKE on MacOS, only Linux.

content/rke/latest/en/installation/_index.md
```diff
@@ -8,6 +8,7 @@ RKE is a fast, versatile Kubernetes installer that you can use to install Kubern
 
 1. [Download the RKE Binary](#download-the-rke-binary)
     1. [Alternative RKE MacOS X Install - Homebrew](#alternative-rke-macos-x-install---homebrew)
```
emcrisostomo: @catherineluse I created this PR because it seems you already support installing RKE on macOS using Brew. MacPorts is just another package management system.

emcrisostomo:
Thanks @catherineluse. I've opened a conversation to highlight the current sources of the webpage where installation on macOS is described (using Brew).

catherineluse:
The existing instructions about installing RKE with HomeBrew shouldn't be in the docs. I'm going to remove it

emcrisostomo:
Hi @catherineluse, I think there might be a misunderstanding here. `rke` in this case are the CLI tools used to interact with an RKE cluster: I think we're talking about a different use care. Would you please have a look here:

https://github.com/rancher/rke/releases/tag/v0.2.10
https://rancher.com/docs/rke/latest/en/installation/#download-the-rke-binary

Rancher is building and distributing `rke` packages for macOS, Linux and Windows. I understand that supporting Linux is one thing, but allowing users to interact with RKE _from a different_ OS is a completely different thing. That's probably one of the reasons this application is written in Go in the first place: OS portability.

catherineluse:
You are right, I got it confused. The RKE CLI can run on MacOS, Linux, or Windows, but can only install Kubernetes on Linux. I will merge the PR.

emcrisostomo:
Thanks @catherineluse!

PR #2334, opened 2020-03-10
Adds section on deploying istio with PSP enabled

cbron:
Issue: https://github.com/rancher/rancher/issues/22469

This issue is technically in 2.4.1, but its not version specific. Should work all the way back to 2.3.


catherineluse:
Lgtm

PR #2350, opened 2020-03-13
updates for security documentation

paraglade:


content/rancher/v2.x/en/security/benchmark-2.3/_index.md
```diff
@@ -34,7 +34,6 @@ When performing the tests, you will need access to the Docker command line on th
 The following scored controls do not currently pass, and Rancher Labs is working towards addressing these through future enhancements to the product.
 
 - 1.1.21 - Ensure that the `--kubelet-certificate-authority` argument is set as appropriate (Scored)
```
sowmyav27: This test will pass when the following is set in cluster.yml . In permissive profile, this test is skipped. So do we add this test to the list of failures ? 
```
services:
  kubelet:
    generate_serving_certificate: true
```

paraglade: updated docs here: https://github.com/rancher/docs/pull/2350/commits/4373346bdcbb0fcd143d33b6fff87043076228a5

PR #2354, opened 2020-03-16
Add Azure Quickstart

nikkelma:


nikkelma:
Depends on rancher/quickstart#57

PR #2355, opened 2020-03-16
Add Google Cloud Quickstart

nikkelma:


nikkelma:
Depends on rancher/quickstart#58

PR #2365, opened 2020-03-19
reflect PagerDuty nomenclature in notifier docs

dacodas:
a [webhook in PagerDuty](https://support.pagerduty.com/docs/webhooks) refers to pushes from PagerDuty to arbitrary HTTP endpoints, not from client machines able to make outbound HTTP connections to PagerDuty's API


dacodas:
My apologies, I did not understand that Rancher uses the Prometheus PagerDuty integration as opposed to the plain v2 Events API. This pull request is incorrect.

PR #2394, opened 2020-03-27
inprove view of code example of nginx proxy

SimonHeimberg:
indent all code in example, only some was before
remove double code end


SimonHeimberg:
reason: the code on https://rancher.com/docs/rancher/v2.x/en/installation/k8s-install/create-nodes-lb/nginx/ looks strange.

PR #2444, opened 2020-04-11
[DOCS UPDATE] Update _index.md for installation page of RKE

developer-guy:
Alternative RKE MacOS X Install - Homebrew  --> this item in list had a broken internal link , so fix it.


developer-guy:
tested with opening page with this link as follows :
https://rancher.com/docs/rke/latest/en/installation/#alternative-rke-macos-x-install---homebrew 

then tested again with the link that after the remove two dashes before homebrew word as follows:
https://rancher.com/docs/rke/latest/en/installation/#alternative-rke-macos-x-install-homebrew

PR #2449, opened 2020-04-14
Docs on how to create OPA Gatekeeper constraints using default templates

prachidamle:
@catherineluse I was not able to load this page and check how it appears; but please review the content. I will try to test in the meantime.


prachidamle:
@catherineluse I was able to test it; 
- the ```yml tag shows and 
- not sure how  to make this link within the page work - [this section below](./Using-the-Default-Constraint-Templates)

PR #2462, opened 2020-04-18
Add bullet to supply --version flag for specific version during install

dkeightley:


catherineluse:
Can you please make the same change in the air gap docs for Kubernetes insalls?

PR #2463, opened 2020-04-20
update the helm rancher example

AlphaWong:
Hi,

I find the example states in the helm instance section with some issues.
I update the helm installation example in order to fit the condition.

# reference
https://github.com/rancher/rancher/blob/master/chart/templates/issuer-letsEncrypt.yaml#L4

Best,
Alpha


catherineluse:
@SheilaghM 
If this example Helm command changes, we would probably need to make the change in many places of the docs.

The outstanding question is, should we show that the cert manager version should be specified in Helm commands?

AlphaWong:
> @SheilaghM
> If this example Helm command changes, we would probably need to make the change in many places of the docs.
> 
> The outstanding question is, should we show that the cert manager version should be specified in Helm commands?

The point is that if you do not specify the run-able version. it will hit this logic https://github.com/rancher/rancher/blob/master/chart/templates/issuer-letsEncrypt.yaml#L4

PR #2468, opened 2020-04-22
Clarify configuration of the kubeconfig path

janeczku:
Configuration of the kubeconfig path is not required for the embedded kubectl command shipped in k3s,.


erikwilson:
LGTM

PR #2499, opened 2020-05-07
Update deprecated option --no-deploy

abmurksi:
Replace old --no-deploy value with --disable to disable Traefik or other options


davidnuzik:
Including @cjellick as a reviewer. He is out today.
I had a GH issue for tracking this and it was closed so wasn't sure if there was some reason we would not want to document this.

cjellick:
this is fine

PR #2502, opened 2020-05-08
Add example for s3 endpoint in specific aws region

Sheldonwl:
It's unclear how to add an endpoint for an S3 bucket that's in a specific region.


Sheldonwl:
Never mind, see that there's a --region parameter.

PR #2506, opened 2020-05-12
Update NLB setup doc for K3s Rancher installs

catherineluse:
This PR is intended to fix problems described in this comment https://github.com/rancher/rancher/issues/26977#issuecomment-626409213, which is related to this issue https://github.com/rancher/rancher/issues/26977

 Currently, our docs for setting up an NLB assume that you have an NGINX ingress and it says "The NGINX ingress controller on the nodes will make sure that port 80 gets redirected to port 443." I'm trying to confirm if this applies to Traefik or if the target groups need to be different.

### DIFF
--- a/content/k3s/latest/en/installation/installation-requirements/_index.md
+++ b/content/k3s/latest/en/installation/installation-requirements/_index.md
@@ -11,18 +11,23 @@ Whether you're configuring a K3s cluster to run in a Docker or Kubernetes setup,
 
 ## Prerequisites
 
-*    Two nodes cannot have the same hostname. If all your nodes have the same hostname, use the `--with-node-id` option to append a random suffix for each node, or otherwise devise a unique name to pass with `--node-name` or `$K3S_NODE_NAME` for each node you add to the cluster.
+Two nodes cannot have the same hostname.
+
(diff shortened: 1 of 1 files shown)


catherineluse:
I'm going to update this PR according to Bastian's comment:

> The traefik ingress controller listens on port 80 and 443 as well. Traffic to port 80 will also automatically be redirected to port 443.
> So it is correct that theLB port 80 should point to ingress controller port 80 and LB port 443 should point to ingress controller port 443.
> The only difference to the nginx ingress controller is, that the health check in our NLB docs does not work like this. The documented health checks do a HTTP request to /healthz . Because the LB health check does not send a host header this will be routed to the default backend of the ingress controller. The default backend of the nginx ingress controller has a /healthz  endpoint, the default backend of traefik has not.
> Since the NLB is working as a Layer 4 TCP loadbalancer anyways, I think it would be correct to also only do a Layer 4 TCP health check to the port of the target group and not an HTTP health check.

jiaqiluo:
LGTM

catherineluse:
Closing this PR in favor of another one that I will make soon

PR #2507, opened 2020-05-12
Explain killall script in K3s docs

catherineluse:
This PR is intended to address this issue https://github.com/rancher/k3s/issues/1238

### DIFF
--- a/content/k3s/latest/en/installation/install-options/killall/_index.md
+++ b/content/k3s/latest/en/installation/install-options/killall/_index.md
@@ -1,0 +1,16 @@
+---
+title: The k3s-killall.sh script
+weight: 4
+---
+
+To allow high availability during upgrades, the K3s containers continue running when the K3s service is stopped.
+
+To stop all of the K3s containers and reset the containerd state, the `k3s-killall.sh` script can be used.
+
+The killall script cleans up containers, K3s directories, and networking components while also removing the iptables chain with all the associated rules. The cluster data will not be deleted.
+
+To run the killall script from a server node, run:
+
+```
+/usr/local/bin/k3s-killall.sh
+```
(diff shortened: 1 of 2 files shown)


davidnuzik:
@catherineluse have a question on:
> "...and networking components while also removing the iptables chain with all the associated rules"

Were you able to verify this with someone? I spoke to Shylaja earlier today (Sorry forgot to update the GH issue with a comment, I have done that now) and we saw we can get k3s to run again just by starting service back up - not sure on the chain but if we got feedback from a dev or someone that works. Just wanted to double-check this is all

content/k3s/latest/en/installation/install-options/restart/_index.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Restarting K3s with Different Installation Options
+weight: 4
+---
+
+If you need to restart K3s with different installation options or parameters, run the `k3s-killall.sh` script, then re-run the installation script with the desired options.
```
erikwilson, line 6: Is this needed?

catherineluse:
@davidnuzik That sentence came from Shylaja when I asked her about it. @erikwilson can you confirm that it's true?

content/k3s/latest/en/installation/install-options/restart/_index.md
catherineluse: I thought it would be needed for a less technical person to understand why the killall script should be used.

davidnuzik:
Okay thanks

content/k3s/latest/en/installation/install-options/restart/_index.md
erikwilson: afaik shouldn't need to

catherineluse: it shouldn't need to be used? why not?

catherineluse:
Shylaja found this [comment that Erik wrote](https://github.com/rancher/k3s/issues/491#issuecomment-495254892) in 2019:

> A k3s-killall.sh script will also be installed when using the install script, which will clean up containers, directories, and networking. The service can then be started again as normal.

ShylajaDevadiga:
LGTM

PR #2517, opened 2020-05-20
jetstack cert-manager doc update

asloan7:
jetstack made a path change, updating doc, also to latest version
reference: https://hub.helm.sh/charts/jetstack/cert-manager


bashofmann:
I can verify that Rancher works well with cert-manager 0.15.0.

PR #2523, opened 2020-05-21
Improve docs for running K3s with Docker

catherineluse:
This PR is intended to address this issue: https://github.com/rancher/k3s/issues/1582

### DIFF
--- a/content/k3s/latest/en/advanced/_index.md
+++ b/content/k3s/latest/en/advanced/_index.md
@@ -113,7 +113,7 @@ K3s will generate config.toml for containerd in `/var/lib/rancher/k3s/agent/etc/
 
 For advanced customization for this file you can create another file called `config.toml.tmpl` in the same directory and it will be used instead.
 
-The `config.toml.tmpl` will be treated as a Golang template file, and the `config.Node` structure is being passed to the template. The following is an example on how to use the structure to customize the configuration file https://github.com/rancher/k3s/blob/master/pkg/agent/templates/templates.go#L16-L32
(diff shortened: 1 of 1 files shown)


content/k3s/latest/en/advanced/_index.md
```diff
@@ -27,11 +28,92 @@ For information about deploying Helm charts, refer to the section about [Helm.](
...
+rancher/pause                    3.1                 da86e6ba6ca19       742kB
+```
+
+# Configuring containerd
+
+K3s will generate config.toml for containerd in `/var/lib/rancher/k3s/agent/etc/containerd/config.toml`.
+
+For advanced customization for this file you can create another file called `config.toml.tmpl` in the same directory and it will be used instead.
+
+The `config.toml.tmpl` will be treated as a Golang template file, and the `config.Node` structure is being passed to the template. The following is an example on how to use the structure to customize the configuration file https://github.com/rancher/k3s/blob/master/pkg/agent/templates/templates.go#L16-L32
```
briandowns, line 116:
```suggestion
The `config.toml.tmpl` will be treated as a Go template file, and the `config.Node` structure is being passed to the template. The following is an example on how to use the structure to customize the configuration file https://github.com/rancher/k3s/blob/master/pkg/agent/templates/templates.go#L16-L32
```

briandowns, line 116: Would we want to create a clickable link like we have above?

catherineluse: I made a clickable link

PR #2546, opened 2020-06-05
EIO-4: corrections needed to address issues found in 5.1.5 tests

paraglade:


catherineluse:
> - Rancher **exec shell** and **view logs** for pods are **not** functional in a cis 1.5 hardened setup when only public ip is provided when registering custom nodes.
> - When setting the `default_pod_security_policy_template_id:` to `restricted` Rancher creates **RoleBindings** and **ClusterRoleBindings** on the default service accounts. The default service accounts should be configured such that it does not provide a service account token and does not have any explicit rights assignments.

On the first bullet item, is there a workaround for that?

On the second item, is that saying that you would expect the setting to be restricted when you change the configuration, but that's not what actually happens?

Also this is minor but I think IP and CIS should be capitalized throughout.

PR #2568, opened 2020-06-23
remove cta

jrcreative:


jrcreative:
@deniseschannon or @catherineluse this is ready for review/merge

jrcreative:
In case this is confusing, this PR is to remove the old CTA that was on the main index page only. The new CTA is still in place.

PR #2578, opened 2020-06-30
Use the installCRDs option for cert-manager

wilsonge:
Uses this feature: https://github.com/jetstack/cert-manager/pull/2775

Given the recommended version in the docs is now 0.15.0 there's no real reason we can't use the flag provided. One less command for people to run too :)

### DIFF
--- a/content/rancher/v2.x/en/installation/k8s-install/helm-rancher/_index.md
+++ b/content/rancher/v2.x/en/installation/k8s-install/helm-rancher/_index.md
@@ -88,9 +88,6 @@ This step is only required to use certificates issued by Rancher's generated CA 
 These instructions are adapted from the [official cert-manager documentation](https://cert-manager.io/docs/installation/kubernetes/#installing-with-helm).
 
 ```
-# Install the CustomResourceDefinition resources separately
-kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v0.15.0/cert-manager.crds.yaml
-
 # **Important:**
 # If you are running Kubernetes v1.15 or below, you
 # will need to add the `--validate=false` flag to your
@@ -114,7 +111,8 @@ helm repo update
 helm install \
   cert-manager jetstack/cert-manager \
   --namespace cert-manager \
-  --version v0.15.0
+  --version v0.15.0 \
+  --set installCRDs=true
 ```
 
 Once you’ve installed cert-manager, you can verify it is deployed correctly by checking the cert-manager namespace for running pods:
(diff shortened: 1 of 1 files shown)


rawmind0 (commented):
@wilsonge , thanks for the PR. There are reasons to not use intallCRDs option as default. It is NOT yet recommended by the provider. It also can cause distinct issues, upgrading or reinstalling the software:
* [upgrading-0.14-0.15](https://cert-manager.io/docs/installation/upgrading/upgrading-0.14-0.15/#new-installcrds-addition) (main reason)`Note: If enabled, when uninstalling, CRD resources will be deleted causing all installed custom resources to be DELETED`
* [upgrading-0.15-0.16](https://cert-manager.io/docs/installation/upgrading/upgrading-0.15-0.16/) `Helm users who use installCRDs=true should not upgrade to v0.16 until a Helm version built upon Kubernetes v1.19 or v1.18.8 is released. At time of writing a PR has been opened for this but is not yet released. This issue only affects v3.1+ however, we don’t currently advise using installCRDs=true`

Due to these reasons, and finally it's just execute an extra command, i think that at the moment, is safer to still recommend the use of `installCRDs=false` on cert-manager installation.

wilsonge:
> upgrading-0.15-0.16 Helm users who use installCRDs=true should not upgrade to v0.16 until a Helm version built upon Kubernetes v1.19 or v1.18.8 is released. At time of writing a PR has been opened for this but is not yet released. This issue only affects v3.1+ however, we don’t currently advise using installCRDs=true

So i mean this one has been patched (REF: https://github.com/jetstack/cert-manager/issues/3121) although the helm release of 3.3.1 is still pending (although no issues have been added to the milestone for 13 days)

> upgrading-0.14-0.15 (main reason)Note: If enabled, when uninstalling, CRD resources will be deleted causing all installed custom resources to be DELETED

I'm unconvinced whether i'd prefer this or not - if you don't have the manager around but still have certificate resources whenever those certificates expire you're going to be real real sad with no upgrade mechanism. Better a swift subtle explosion than a slow grind imo :P 

But I guess it's your guys call in this case as it's your docs though :) So if i haven't convinced you say the word and I'll go back to the custom resource (but I'd like to document why we're not recommending using the option at this point - because it wasn't obvious to me may as well make it explicit)

nikkelma:
We've had some similar discussion in [a quickstart PR, too](https://github.com/rancher/quickstart/pull/106#issuecomment-670761506).

There's more than just the uninstall case that's impacted. If an upgrade of cert-manager fails and for some reason a full uninstall/reinstall cycle is performed, `installCRDs=true` would wipe out all custom resources even though the intention was to preserve all cert-manager functionality. That's much more dangerous to me than leaving controller-less custom resources around.

rawmind0:
@wilsonge , your arguments didn't convinced me. `installCRDs=false` option is safer than the other one, at the cost of execute just 1 extra command on installation. At the moment, we should keep recommending this one.

We are updating/testing Rancher chart to support cert-manager v1.0.x, rancher/rancher#29017 . Could you please move PR to v1.0.1 instead of v0.15.0??

wilsonge:
> your arguments didn't convinced me

I tried :) 

All done

rawmind0 (approved):
LGTM

bashofmann:
@wilsonge now that 2.5 has been releases, the docs structure changed a bit. Could you rebase this PR again? And maybe when you are at it, directly update it to 1.0.3 :-)

PR #2579, opened 2020-06-30
Update airgapped installation documentation to newer version of cert manager

wilsonge:
- Update docs to cert manager 0.15.1 (currently non-airgapped docs are using 0.15.0)
- Use the installCRDs option which allows us to skip some steps

### DIFF
--- a/content/rancher/v2.x/en/installation/other-installation-methods/air-gap/install-rancher/_index.md
+++ b/content/rancher/v2.x/en/installation/other-installation-methods/air-gap/install-rancher/_index.md
@@ -85,22 +85,19 @@ By default, Rancher generates a CA and uses cert-manager to issue the certificat
 1. Fetch the latest cert-manager chart available from the [Helm chart repository](https://hub.helm.sh/charts/jetstack/cert-manager).
 
    ```plain
-   helm fetch jetstack/cert-manager --version v0.12.0
+   helm fetch jetstack/cert-manager --version v0.15.1
    ```
 
 1. Render the cert manager template with the options you would like to use to install the chart. Remember to set the `image.repository` option to pull the image from your private registry. This will create a `cert-manager` directory with the Kubernetes manifest files.
    ```plain
-   helm template cert-manager ./cert-manager-v0.12.0.tgz --output-dir . \
+   helm template cert-manager ./cert-manager-v0.15.1.tgz --output-dir . \
        --namespace cert-manager \
        --set image.repository=<REGISTRY.YOURDOMAIN.COM:PORT>/quay.io/jetstack/cert-manager-controller \
        --set webhook.image.repository=<REGISTRY.YOURDOMAIN.COM:PORT>/quay.io/jetstack/cert-manager-webhook \
        --set cainjector.image.repository=<REGISTRY.YOURDOMAIN.COM:PORT>/quay.io/jetstack/cert-manager-cainjector
+       --set installCRDs=true
    ```
 
-1. Download the required CRD file for cert-manager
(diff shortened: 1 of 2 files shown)


rawmind0 (commented):
Upgrade airgapped install docs to user newer `cert-manager` version, lgtm.

At the moment, i disagree in the use of `installCRDs=true` as recommended option for installing `cert-manager`. It could generate possible issues upgrading or reinstalling the software, to just save one extra command line. Please, take a look at the comment, https://github.com/rancher/docs/pull/2578#pullrequestreview-476689901

wilsonge:
> At the moment, i disagree in the use of installCRDs=true as recommended option for installing cert-manager. It could generate possible issues upgrading or reinstalling the software, to just save one extra command line. Please, take a look at the comment, #2578 (review)

OK will discuss whether or not to use it there - but assume in the meantime for the purposes of this PR we need to move from `00-crds.yaml` to `cert-manager.crds.yaml` as in those docs? (as obviously 00-crds.yaml is removed in v0.15).

> Upgrade airgapped install docs to user newer cert-manager version, lgtm.

Given it's now out are you happy with me moving both to 0.16 (I haven't tested this one in the airgap yet - so will need directions from you guys if you've tested or not)

rawmind0:
> OK will discuss whether or not to use it there - but assume in the meantime for the purposes of this PR we need to move from 00-crds.yaml to cert-manager.crds.yaml as in those docs? (as obviously 00-crds.yaml is removed in v0.15).

Obviously, it should refer to the proper crd file, v1.0.1, https://github.com/jetstack/cert-manager/releases/download/v1.0.1/cert-manager.yaml


> Given it's now out are you happy with me moving both to 0.16.1 (I haven't tested this one in the airgap yet - currently running 0.15.1 - so will need directions from you guys if you've tested or not)

We are updating/testing Rancher chart to support cert-manager v1.0.x, https://github.com/rancher/rancher/pull/29017 . Could you please move both to v1.0.1 instead of v0.16.1??

wilsonge:
Done

wilsonge:
@rawmind0 On testing for airgapped - v1 of cert manager doesn't work - because https://github.com/rancher/rancher/blob/master/chart/templates/issuer-letsEncrypt.yaml#L4 is checking that the version is equal to 0.x (which isn't true) therefore specifies the incorrect apiVersion of alpha1 instead of alpha2

rawmind0:
@wilsonge thanks for the quick update on that. 

> @rawmind0 On testing for airgapped - v1 of cert manager doesn't work - because https://github.com/rancher/rancher/blob/master/chart/templates/issuer-letsEncrypt.yaml#L4 is checking that the version is equal to 0.x (which isn't true) therefore specifies the incorrect apiVersion of alpha1 instead of alpha2

I already know that Rancher installations (airgapped or not) using cert-manager v1.0.0 doesn't work at the moment. As mentioned, due to that reason, PR https://github.com/rancher/rancher/pull/29017 has been submitted to address the issue. Until PR is not merged, v1 testing should be done using PR content

wilsonge:
Cool so this shouldn't be merged until rancher/rancher#29017 is merged and in a tagged release of the helm chart

rawmind0 (approved):
PR https://github.com/rancher/rancher/pull/29017 already merged

LGTM

catherineluse:
I changed the version to v1.0.4 https://github.com/rancher/docs/commit/ec87dc9dcb794f1454d6ad8d629a2ea4a2662b9f per the discussion here https://github.com/rancher/docs/issues/2803

Issue #2577, opened 2020-06-30
Cluster Administration Table links to 404 pages and duplications of the same contents in the bottom section of this page.


yzeng25:
URL: https://rancher.com/docs/rancher/v2.x/en/cluster-admin/

Issue 1: Cluster Administration Table links to 404 pages. The first column of the table on this page are links, but all of them fail to link to valid pages.

Issue 2: Duplications of the same contents in the bottom section of this page. The same contents of table appear in the bottom section of this page without correct formatting.

superseb:
I guess this is related to https://github.com/rancher/docs/pull/2571, @bashofmann can you check?

bashofmann:
Will look at it asap

bashofmann:
@superseb https://github.com/rancher/docs/pull/2585 fixes both issues for me. Learnt a bit about how hugo page bundles work.

PR #2585, opened 2020-07-03
Move cluster-capabilities-table to a headless bundle

bashofmann:
Move cluster-capabilities-table to a headless bundle and include it as a page instead of reading the file

This prevents the table to automatically be included on pages of other bundles.
By including the page instead of reading the file also the shortcodes in the table are expanded correctly.

This fixes broken links in the table.

Fixes https://github.com/rancher/docs/issues/2577


catherineluse:
@bashofmann I accidentally merged this too soon. After merging these changes I don't see the tables. Were you able to see the tables if you build the docs locally with `./scripts/dev'?

bashofmann:
@catherineluse There seems to be a difference between doing a fresh build and what the watch process is doing when you change something. It was definitely showing up while I was working on it, but when doing a fresh built it's not working. I'm very sorry for that :(

I also figured out the reason, you can only use `.Site.GetPage` on a complete page bundle and include the `index.md` and not on a single file, at least on a fresh build.

I opened https://github.com/rancher/docs/pull/2592 which fixes this for me.

catherineluse:
Thank you @bashofmann looks good now.

PR #2618, opened 2020-07-28
Replaced 'alpine' image with Murali's swiss-army-knife

Tejeev:


catherineluse:
What's the advantage of Swiss army knife compared to Alpine?

PR #2619, opened 2020-07-28
synchronised and updated overlay test

Tejeev:
RKE and Rancher had different versions of the test.  I synced them and changed them so that they utilize Murali's swiss-army-knife which allows much more freedom for troubleshooting.


Tejeev:
@catherineluse the advantage of swiss-army-knife is that it has many many more tools we use when troubleshooting networks and we often use this test, but with the s-a-k image to troubleshoot for customers, so we can use the deployed images to further investigate root causes once we've tested the general ping ICMP communication of the overlay network

Tejeev:
Murali and I are making changes to the test in https://github.com/rancher/docs/pull/2620.  When they are approved, I'll merge them here as well.
Would be good if we could have one source of truth for this that is pulled to both docs, instead of two different docs with the same test

catherineluse:
Closing this PR because after talking to TJ I think https://github.com/rancher/docs/pull/2620/files covers it. The RKE troubleshooting guide can point to that test.

PR #2625, opened 2020-07-31
Improve Keycloak authentication documentation

bashofmann:
* Explain in text and with screenshots how to correctly configure the Rancher SAML client in Keycloak, including mappers
* Provide examples for field configurations in Rancher that map to the Keycloak defaults
* Move the note on how to get a metadata.xml file with Keycloak >= 6.0.0 directly to the Prerequisites so that it is hard to miss


bashofmann:
I tested that everything works and made the screenshots myself.

Issue #2613, opened 2020-07-20
The RKE/k3s kubeconfig should be used for Rancher helm upgrades


dkeightley:
A kubeconfig for the local cluster can be generated from the Rancher UI. If this is mistakenly used for helm upgrades, the dependency on the Rancher API being available can cause issues - as Rancher pods will be likely re-reployed and prevent the apply of other objects.

An upgrade with this kubeconfig resulted in the below error:
`Error: UPGRADE FAILED: cannot patch "rancher" with kind Ingress: an error on the server ("<html>\r\n<head><title>502                                    Bad Gateway</title></head>\r\n<body>\r\n<center><h1>502 Bad Gateway</h1></center>\r\n<hr><center>openresty/1.15.8.1</center>\r\n</body>\r\n</html>") has prevented the request from succeeding (patch ingresses.extensions rancher)`

To prevent this, stating a requirement for upgrades to use the RKE/k3s kubeconfig, which performs a direct kube-apiserver connection.

superseb:
We can possibly re-use this part from migrating docs: https://rancher.com/docs/rancher/v2.x/en/upgrades/upgrades/migrating-from-rke-add-on/#point-kubectl-at-your-rancher-cluster

PR #2635, opened 2020-08-18
Confirm the correct kubeconfig is being used

dkeightley:
https://github.com/rancher/docs/issues/2613


content/rancher/v2.x/en/upgrades/upgrades/ha/_index.md
```diff
@@ -22,6 +22,28 @@ If you installed Rancher using the RKE Add-on yaml, follow the directions to [mi
...
+
+```
+export KUBECONFIG=$(pwd)/kube_config_rancher-cluster.yml
+```
+
+Verify that it contains the correct `server` parameter, it should point directly to one of your cluster nodes on port `6443`.
+
+```
+kubectl config view -o=jsonpath='{.clusters[*].cluster.server}'
+https://NODE:6443
```
superseb, line 37: Do we want to add a note for the use of `localhost` in case of k3s (where people did not replace it with their load balancer/node address)?

content/rancher/v2.x/en/upgrades/upgrades/ha/_index.md
```diff
@@ -22,6 +22,28 @@ If you installed Rancher using the RKE Add-on yaml, follow the directions to [mi
...
+```
+
+If the output from the command shows your Rancher hostname with the suffix `/k8s/clusters`, the wrong kubeconfig YAML file is configured. It should be the file that was created with RKE or K3s when creating the cluster to run Rancher.
+
+Confirm that Rancher pods are running within the cluster.
```
superseb, line 42: If this is a separate step, should it be in the list above so it doesn't get overlooked?

catherineluse:
@dkeightley Please rebase this PR on master

PR #2637, opened 2020-08-18
Cleanup Rancher port requirements

bashofmann:
Port requirements were documented in 3 places in the Rancher docs. This consolidates the documentation in one place and correctly distinguishes between the Rancher management cluster and downstream clusters.

This also addresses https://github.com/rancher/rancher/issues/27502.


catherineluse:
@sangeethah Is it possible for a QA to review this PR?

bashofmann:
Rebased to resolve the conflict.

catherineluse:
I need to edit the port requirements because we need to add the ports for installing Rancher using RancherD. And that change is for the v2.5.4 release. So I'm going to switch this into staging, then make the change in staging. When it's in staging it will also be easier to review.

PR #2653, opened 2020-08-25
Update _index.md

eino-makitalo:
When  I used offered Graph API https://graph.microsoft.com , Rancher 2.4.5 failed with error code. Changing value https://graph.windows.net/  make it work.


catherineluse:
@SheilaghM 
Before merging this change, we should verify which version of Rancher the API endpoint stopped working in. Then we can say, for version x, use the API endpoint, and for version y, use the URL.

Issue #2662, opened 2020-09-02
Update docs to reflect newly required AWS perms in 2.4.6 for using AWS node templates


rosskirkpat:
There was a breaking change in v2.4.6 regarding the required AWS perms for node templates (https://github.com/rancher/rancher/issues/27965)

Our docs do not list the proper IAM KMS permissions for the Amazon node driver required for this 2.4.6 change https://rancher.com/docs/rancher/v2.x/en/cluster-provisioning/rke-clusters/node-pools/ec2/

We should also note this as a breaking change. On 2.4.5, you would be able to create/edit/clone AWS node templates but once upgrading to 2.4.6 or 2.4.7, you would now be unable to if you are missing the `kms:ListKeys` permission. 

There is also an open issue regarding the UX encountered when you don't have the correct AWS perms and try to create/edit/clone an AWS node template: https://github.com/rancher/rancher/issues/28719

We should consider calling out that if you are stuck on the `Loading` stage after hitting `Next: Authenticate & configure nodes`, you should enable the dev tools for your browser to see if there is an error present. 

gz#12099

catherineluse:
Since this is no longer an issue in v2.4.8+ I think these notes should address it. https://github.com/rancher/docs/pull/2673

PR #2673, opened 2020-09-04
Document permissions for EC2 node templates in v2.4.6 and v2.4.7

catherineluse:
Addresses https://github.com/rancher/docs/issues/2662


sowmyav27 (approved):
LGTM

PR #2675, opened 2020-09-05
Update docker install command for Rancher 2.5

catherineluse:


content/rancher/v2.x/en/upgrades/upgrades/single-node/_index.md
```diff
@@ -0,0 +1,332 @@
+---
```
cbron, line 1: Is this whole file related, and should I review it ?

content/rancher/v2.x/en/installation/other-installation-methods/air-gap/install-rancher/_index.md
```diff
@@ -12,6 +12,10 @@ This section is about how to deploy Rancher for your air gapped environment. An
 
 > **Note:** These installation instructions assume you are using Helm 3. For migration of installs started with Helm 2, refer to the official [Helm 2 to 3 migration docs.](https://helm.sh/blog/migrate-from-helm-v2-to-helm-v3/) This [section]({{<baseurl>}}/rancher/v2.x/en/installation/options/air-gap-helm2) provides a copy of the older air gap installation instructions for Rancher installed on Kubernetes with Helm 2, and it is intended to be used if upgrading to Helm 3 is not feasible.
 
+### Privileged Access for Rancher v2.5+
+
+When the Rancher server is deployed in the Docker container, a local Kubernetes cluster is installed within the container, then Rancher is deployed on the cluster. Because privileged access is required to run containers within containers, you will need to install Rancher with the `--privileged` option in order to access many features of Rancher, including deploying pods on the local Kubernetes cluster.
```
cbron, line 17:
```suggestion
When the Rancher server is deployed in the Docker container, a local Kubernetes cluster is installed within the container for Rancher to use. Because many features of Rancher run as deployments, and privileged mode is required to run containers within containers, you will need to install Rancher with the `--privileged` option.
```

Feel free to refactor the grammar, but the previous way suggested that it was possible to run rancher without privileged and some features just wouldn't run, but Rancher won't run at all without the privileged flag now.

cbron (changes requested):

cbron:
Oh and also we need to do the `addLocal` helm option changes, where auto goes away in favor of just true or false.

content/rancher/v2.x/en/upgrades/upgrades/single-node/_index.md
catherineluse: This file was supposed to be deleted because it's now located at /rancher/v2.x/en/installation/other-installation-methods/single-node-docker/single-node-upgrades/. I have added a commit to remove this duplicate.

catherineluse:
I have added a commit to change the default value for addLocal to true in the Helm chart config reference.

cbron (commented):
This looks good, but we do need to add one more thing. Basically that if you turn addLocal off then most v2 features just won't work, including EKS provisioner. Example: https://github.com/rancher/rancher/issues/28764

catherineluse:
Added the note

> If you turn addLocal off, most Rancher v2.5 features won't work, including the EKS provisioner.

cbron (approved):
lgtm

PR #2689, opened 2020-09-16
I used helm2 so maybe these are not needed in helm3

StefanIGit:


catherineluse:
It looks like this PR is changing Helm 3 commands to Helm 2 commands. This shouldn't be necessary because the older version of the docs with the Helm 2 commands are retained here: https://rancher.com/docs/rancher/v2.x/en/installation/options/helm2/

PR #2705, opened 2020-09-28
Eks syncing

rmweir:
Added documentation for syncing. The first paragraph condenses the most import points. The remainder explains how syncing works at a high level.

**Issue:**
https://github.com/rancher/rancher/issues/26930

### DIFF
--- a/content/rancher/v2.x/en/cluster-provisioning/hosted-kubernetes-clusters/eks/_index.md
+++ b/content/rancher/v2.x/en/cluster-provisioning/hosted-kubernetes-clusters/eks/_index.md
@@ -93,6 +93,8 @@ The following capabilities have been added for configuring EKS clusters in Ranch
 - Add restrictions to public access
 - Use your cloud credentials to create the EKS cluster instead of passing in your access key and secret key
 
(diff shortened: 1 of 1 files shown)


catherineluse:
Hi Ricky,

This information looks pretty good to me. I think we could frame it so that the prescriptive guidance stands out a little more.

- In the cluster configuration reference, after the list of new features but before the tabs, I think there should be a sentence that says something along the lines of "Due to the way that the cluster data is synced with EKS, if the cluster is modified in the EKS console and in Rancher within five minutes, it could cause some changes to be overwritten. For information about how the sync works and how to configure it, refer to [this section.](#syncing)" So it could prevent surprises.
- In the Troubleshooting section, we should have a sentence that says "If your changes were overwritten, it could be due to the way the cluster data is synced with EKS. Changes shouldn't be made to the cluster in the EKS console and in Rancher within a five-minute span. For information on how this works and how to configure the refresh interval, refer to [Syncing.](#syncing)
- I think the last paragraph should have a subsection title `### Configuring the Refresh Interval` so it's easier to find.

rmweir:
@catherineluse thank you for the great feedback. Made all the recommended changes. Replaced mentions of EKS console with something more general followed by EKS console as an example.

catherineluse (approved):
LGTM

PR #2730, opened 2020-10-03
Update backup restore docs

mrajashree:


content/rancher/v2.x/en/backups/migrating-rancher/_index.md
```diff
@@ -59,10 +71,10 @@ Follow the steps to [install cert-manager]({{<baseurl>}}/rancher/v2.x/en/install
 
 ### 4. Bring up Rancher with Helm
 
+Use the same version of Helm to install Rancher, that was used on the first cluster.
+
 ```
-helm upgrade rancher rancher-alpha/rancher \
-  --version 2.5.0-alpha1 \
+helm install rancher rancher-latest/rancher \
```
mrajashree, line 77: @maggieliu I just want to confirm that the 2.5 version will be tagged as rancher-latest right? so this helm install command will install rancher 2.5?

mrajashree: confirmed offline that this is what the command should be, and after 2.5 is released it will be tagged latest

cloudnautique:
We should note that the backup operator should only be installed in the local cluster. I also think it is a good idea to call out that it’s s only backing up the Rancher management workload.

mrajashree:
@cloudnautique The chart is hidden from downstream clusters, and there's a note in the main backups page that it it to be installed only in the local cluster. Also since the docs call it rancher backup, I think that will make it clear that it's only backing up the rancher app

catherineluse (approved):
LGTM

PR #2733, opened 2020-10-04
Reword istio overview page, make legacy docs more obvious...

brendarearden:
Format resources table, Change additional scrape config
information to match current monitoring default

### DIFF
--- a/content/rancher/v2.x/en/istio/rbac/_index.md
+++ b/content/rancher/v2.x/en/istio/rbac/_index.md
@@ -9,7 +9,7 @@ This section describes the permissions required to access Istio features.
 
 The rancher istio chart installs three `ClusterRoles` 
 
-# Cluster-Admin Access
+## Cluster-Admin Access
 
 By default, only those with the `cluster-admin` `ClusterRole` can:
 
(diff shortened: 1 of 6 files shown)


catherineluse (changes requested):
@brendarearden can you please replace the placeholder URLs?

content/rancher/v2.x/en/istio/rbac/_index.md
```diff
@@ -26,21 +26,21 @@ By default, only Admin and Edit roles can:
...
 
-| ClusterRole create by chart | Default K8s ClusterRole  | Rancher Role | 
-| ------------------------------| ---------------------------|---------|
-| `istio-admin` | admin| Project Owner, Project Member | 
-| `istio-edit`| edit | Project Owner, Project Member | 
-| `istio-view` | view | Read-only | 
+ClusterRole create by chart | Default K8s ClusterRole  | Rancher Role |
+ ------------------------------:| ---------------------------:|---------:|
+ `istio-admin` | admin| Project Owner, Project Member | 
+ `istio-edit`| edit | Project Owner, Project Member |
```
cbron, line 36: I believe this is incorrect, although it looks like they get same permissions. 
project-member would be edit: https://github.com/rancher/charts/blob/dev-v2.5-source/packages/rancher-istio/charts/templates/edit-role.yaml#L6
project-owner would be admin: https://github.com/rancher/charts/blob/dev-v2.5-source/packages/rancher-istio/charts/templates/admin-role.yaml#L6

Let me know if you found different. But the two roles do have the same permissions.

brendarearden: They do get the same permissions, so that is why I listed the roles that way, but you are right that it is confusing. I went ahead and updated istio-admin to be project owner, and then istio-edit to be project member

catherineluse (approved):
LGTM

PR #2735, opened 2020-10-05
Respond to feedback on CIS scan docs

catherineluse:
For this this PR:

- I added the information about the new columns in the report
- I validated the test descriptions against the 1.5.0 benchmark (they didn't change)
- On the configuration page it now says the default profiles and benchmarks should not be edited. (It already said that in the benchmark configuration section, but now it also says that in the profile configuration section.)
- It now says that all the tests that are skipped and NA - all will be counted as Not Applicable in the v2.5 report. The skipped test count will only mention the user-defined skipped tests.
- The RBAC section says that in Rancher, only cluster-owner and global-admin have admin access by default.


prachidamle:
lgtm

PR #2712, opened 2020-09-30
RBAC for CIS v2

prachidamle:

PR #2742, opened 2020-10-05
Edit CIS RBAC docs

prachidamle:
Editing as per the discussion here https://github.com/rancher/docs/pull/2712#discussion_r499883395


catherineluse (approved):
LGTM

cbron (approved):
This works for now. Possible that the people might be confused by the table, if they assume they are namespaced. But paragraph should clear up. We'll see if we get questions.

PR #2744, opened 2020-10-06
Added note that addLocal not an option in v2.5.0.

dnoland1:


dramich (approved):

catherineluse:
can we add the helm chart option to the table?

example: restrictedAdmin=true
default value: false
Description: When this option is set to true, the initial Rancher user has restricted access to the local Kubernetes cluster to prevent privilege escalation. For more information, see the section about the [restricted-admin role.]({{<baseurl>}}/rancher/v2.x/en/admin-settings/rbac/global-permissions/#restricted-admin)

the link will work after https://github.com/rancher/docs/pull/2745 is merged

catherineluse:
I've added the Helm chart option here https://github.com/rancher/docs/pull/2746

PR #2746, opened 2020-10-06
Add restrictedAdmin option to Helm chart options

catherineluse:


catherineluse:
I have added a thing that says the restrictedAdmin option is available as of v2.5 https://github.com/rancher/docs/commit/f0c5521edebce47f2a8a3fdd30aa3021f18ed9db

PR #2755, opened 2020-10-06
Update _index.md to correctly reference cloud_provider configuratuion 

ahbosch:
Cloud provider config is under rancher_kubernetes_engine_config and should be noted that way also the folder selection is the name of the folder only, you don't have to specify the datacenter or a "/" before the name. 

The initial configuration of only "cloud_provider" with no indents is misleading and the folder reference is incorrect as I get an error in rancher logs when adding the folder with datacenter name, but works perfectly when only referencing the folder name


catherineluse:
@luthermonson Can you please review this change?

luthermonson (approved):

PR #2768, opened 2020-10-10
Update Istio docs for Network Isolation

cortopy:
In clusters with namespace Network Isolation enabled, the default installation of Istio needs further tweaking for it to work


brendarearden (approved):
looks good!

PR #2774, opened 2020-10-13
Add taints and tolerations section to logging

nickgerace:
Add taints and tolerations section to logging. Include example YAML, and
provide both the overall backdrop, and the Rancher-specific use cases.

### DIFF
--- a/content/rancher/v2.x/en/logging/v2.5/_index.md
+++ b/content/rancher/v2.x/en/logging/v2.5/_index.md
@@ -8,14 +8,15 @@ weight: 1
 - [Configuring the Logging Output for the Rancher Kubernetes Cluster](#configuring-the-logging-output-for-the-rancher-kubernetes-cluster)
 - [Enabling Logging for Rancher Managed Clusters](#enabling-logging-for-rancher-managed-clusters)
 - [Configuring the Logging Application](#configuring-the-logging-application)
+- [Working with Taints and Tolerations](#working-with-taints-and-tolerations)
 
 
 ### Changes in Rancher v2.5
 
 The following changes were introduced to logging in Rancher v2.5:
 
-- Rancher's logging feature is now powered the [Banzai Cloud Logging operator](https://banzaicloud.com/docs/one-eye/logging-operator/) instead of Rancher's in-house logging solution.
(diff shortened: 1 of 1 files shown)


nickgerace:
Removed @catherineluse for the time being to let folks check out the PR 😁

content/rancher/v2.x/en/logging/_index.md
```diff
@@ -242,4 +243,33 @@ spec:
...
+    - key: cattle.io/os
+      operator: "Equal"
+      value: "linux"
+      effect: NoSchedule 
+  nodeSelector:
+    kubernetes.io/os: linux
+```
+
+In the above example, we ensure that our pod only runs on Linux nodes, and we add a *toleration* for the taint we have on all of our Linux nodes.
+You can do the same with Rancher's existing taints, or with your own custom ones.
```
paynejacob, line 275: This is a good overview of the taint's and tolerations but I would expand on it to cover how to pass tolerations to the fluentbit, fluentd, and operator containers. It is also probably worth mentioning why we don't schedule to windows nodes.

paynejacob (changes requested):

content/rancher/v2.x/en/logging/_index.md
nickgerace: I agree. Will do.

content/rancher/v2.x/en/logging/v2.5/_index.md
```diff
@@ -236,4 +237,64 @@ spec:
...
+  nodeSelector:
+    kubernetes.io/os: linux
+```
+
+In the above example, we ensure that our pod only runs on Linux nodes, and we add a ```toleration``` for the taint we have on all of our Linux nodes.
+You can do the same with Rancher's existing taints, or with your own custom ones.
+
+**Why do we not schedule logging-related pods on Windows nodes?**
+
+Not all parts of the logging stack are compatible with Windows Kubernetes nodes.
```
paynejacob, line 277: no parts of the logging stack...

paynejacob (changes requested):

content/rancher/v2.x/en/logging/v2.5/_index.md
nickgerace: Fixed. I did not want to make an assumption in case an image or two offered experimental Windows support, etc. However, I changed the wording.

paynejacob (approved):

nickgerace:
Spoke with Catherine and merged accordingly.

PR #2801, opened 2020-10-27
Document RancherD installation

catherineluse:
1. What are the advantages vs. disadvantages of installing rancherD with `rancherd server` vs systemd? I ask because Craig’s blog post said “You can launch the binary directly via `rancherd server`, but we’re going to stick with the systemd service for this demo.” Edit: we prefer the systemd installation method.
2. When I tried the steps in Craig’s blog post, I installed Rancher successfully, but couldn’t interact with the RKE2 cluster using kubectl. Edit: Needed to log in as root from the beginning.
3. How do you set up rancherd with the fixed registration address? Edit: We should recommend configuring the fixed registration address in the config.yaml at the beginning, even for the initial single-node install.
3. Should we include steps to install rancherd in an air gapped environment? Edit: We're not going to document air gapped RancherD installs at this time because it hasn't finished being tested yet.
4. Does RKE2 have a fixed registration address like in the K3s docs? Edit: yes
5. which do we recommend for most people, manual upgrades or automatic upgrades? Edit: We recommend re-running the installation script.

### DIFF
--- a/content/rancher/v2.x/en/installation/install-rancher-on-linux/rancherd-configuration/_index.md
+++ b/content/rancher/v2.x/en/installation/install-rancher-on-linux/rancherd-configuration/_index.md
@@ -1,0 +1,316 @@
+---
+title: RancherD Configuration Reference
+weight: 1
+---
+
+In RancherD, a server node is defined as a machine (bare-metal or virtual) running the `rancherd server` command. The server runs the Kubernetes API as well as Kubernetes workloads.
+
+An agent node is defined as a machine running the `rancherd agent` command. They don't run the Kubernetes API. To add nodes designated to run your apps and services, join agent nodes to your cluster.
+
+In the RancherD installation instructions, we recommend running three server nodes in the Rancher server cluster. Agent nodes are not required.
+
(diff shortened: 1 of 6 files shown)


content/rancher/v2.x/en/installation/install-rancher-on-linux/rancherd-configuration/_index.md
```diff
@@ -0,0 +1,316 @@
...
+rancherd agent [OPTIONS]
+```
+
+The following options are available.
+
+### Config
+
+| Option | Description |
+|--------|-------------|
+|   `--config FILE, -c FILE`  | Load configuration from FILE (default: "/etc/rancher/rancherd/config.yaml") |
```
ShylajaDevadiga, line 245: Config file location is same as rke2  /etc/rancher/rke2/config.yaml

catherineluse: I have changed it here https://github.com/rancher/docs/pull/2822

PR #2818, opened 2020-11-03
added docs for logging v2.5 migration

ebauman:
Added documentation for migrating from legacy logging to v2.5 logging.


content/rancher/v2.x/en/logging/v2.5/migrating/_index.md
```diff
@@ -0,0 +1,191 @@
...
+   2. You do not need to configure any filters if you do not wish - default behavior does not require their creation
+   3. Define your cluster output(s)
+
+This will result in logs from all sources in the cluster (all pods, and all system components) being collected and sent to the output(s) you defined in the _ClusterFlow_. 
+
+# Project Logging
+
+Logging in v2.5+ is not project-aware. This means that in order to collect logs from pods running in project namespaces, you will need to define _Flows_ for those namespaces. 
+
+To collect logs from a specific namespace, follow these steps:
```
cloudnautique, line 71: There is an alternate approach that I would pitch first, which is to define the application log flow and output as part of your application definition. Assuming the logging operator is deployed in the cluster. This approach could be favored as it allows the application to be portable.

ebauman: I think there's a counterargument to be made here that including those objects in the app manifests actually reduces portability because you cannot guarantee that the banzai logger is present in all clusters. So you get tied to a specific logging implementation, or your application refuses to deploy because those CRDs don't exist in the target cluster.

cloudnautique: OK, I can see not making the recommended path especially when using straight manifests. But, I think we should suggest it, particularly if you are using Helm because you can use helm capabilities detection to determine if a CRD and version exist.

cloudnautique: Thinking about it more, we can hold off on suggesting this for now. It’s something that a user could still do, but we don’t need to include it in the guide.

cloudnautique (approved):

catherineluse:
@ebauman We have a section about installing the logging app here https://rancher.com/docs/rancher/v2.x/en/logging/v2.5/#enabling-logging-for-rancher-managed-clusters. In the section about installing logging, can you please link to the existing section and put the new screenshot there?

ebauman:
@catherineluse I have linked to the existing install instructions as recommended. I removed the image as I felt it was better to use the existing instructions and not include a screenshot that could potentially change.

PR #2824, opened 2020-11-10
add flannel backend port configuration example

chanjarster:
Sometimes the default 8472/udp port will conflict with virtualization platform, and we need to change it, while the doc doesn't mention that.


superseb (changes requested):
Thanks for the PR. I'm fine with adding the keys as example but the comment about Windows only applies to Rancher, not to RKE which this applies to. Also Windows is only supported with Flannel.

chanjarster:
@superseb fixed

superseb (approved):

PR #2828, opened 2020-11-11
Remove Longhorn no longer needed storageclass info

jcleira:
This MR removes a no longer needed step about running a `kubectl create -f` method with Longhorn `storageclass` as that `storageclass` is now included in [the base yaml executed one step before](https://raw.githubusercontent.com/longhorn/longhorn/master/deploy/longhorn.yaml)


yasker (approved):
LGTM. Longhorn has included a storage class in the deployment.

PR #2837, opened 2020-11-12
Update OS requirements section in K3s docs

catherineluse:
This PR is intended to bring the K3s OS requirements inline with the recommendations in this Slack thread. https://rancher.slack.com/archives/C039Y2P93/p1604507461130500?thread_ts=1604459764.129000&cid=C039Y2P93

This should clarify that as far as support goes, the Rancher support and maintenance terms define what is tested and supported. There is no separate support and maintenance document for K3s, mainly because support is not sold separately for K3s.


davidnuzik:
Linking to GH issue: https://github.com/rancher/docs/issues/2841

davidnuzik (approved):

Issue #2356, opened 2020-03-16
Node options for imported cluster are incorrect


loganhz:
It says 

```
Notes for Imported Nodes
Although you can deploy workloads to an imported cluster using Rancher, you cannot manage individual cluster nodes. All management of imported cluster nodes must take place outside of Rancher.
```

We can do Cordon Drain Edit and View API for imported node in UI

https://rancher.com/docs/rancher/v2.x/en/cluster-admin/nodes/#node-options-for-each-type-of-cluster

![image](https://user-images.githubusercontent.com/21168270/76782272-02319200-67eb-11ea-8771-50b64c830eae.png)
![image](https://user-images.githubusercontent.com/21168270/76782323-18d7e900-67eb-11ea-93e1-313c012bb84e.png)

PR #2850, opened 2020-11-19
Update table showing capabilities for Rancher managed clusters

catherineluse:
I have updated the table that shows capabilities for different types of Rancher managed clusters. The changes are:

- Removes the imported clusters column
- Adds new columns, one for registered EKS clusters and one for all other registered clusters

I added checkmarks assuming the capabilities for registered EKS clusters are the same as Rancher launched clusters except for deletions, and the column for all other registered clusters is the same as the capabilities for imported clusters in Rancher v2.4.

Screen Shot 2020-11-19 at 12 30 23 AM

Here is the previous table for reference:

Screen Shot 2020-11-19 at 12 38 26 AM


rmweir (commented):
I know we had it before but can you actually manage nodes for a non-EKS registered cluster? I think it depends on how you mean manage node. I don't think you can upgrade them or anything. Maybe upgrade nodes could be another row? Some of our other cloud providers are limited too, but maybe I am not interpreting this correctly.

catherineluse:
https://github.com/rancher/docs/issues/2356

catherineluse:
@deniseschannon Should the 'manage nodes' row be broken down into more detail? And should the table show that the ability to manage nodes is limited for some cluster types?

deniseschannon:
Yes, we should probably clarify what managing nodes means and break it down.

catherineluse:
https://github.com/rancher/docs/issues/2356

PR #2856, opened 2020-11-21
Rancher v2.5 guides with CIS 1.5

deniseschannon:
https://github.com/rancher/docs/issues/2857


deniseschannon:
@catherineluse This is my first pass at trying to separate the existing hardening/self assessment guides into

Rancher version
CIS benchmark version
k8s distro type 

Please edit as needed to make it sound good. :)

catherineluse:
I've edited it. Is it ok to merge?

deniseschannon:
Yes, I think so, we need to add/include CIS 1.6 when we can but publishing CIS 1.5 should also be good.

PR #2906, opened 2020-12-10
Update docs for embedded etcd ports

brandond:
Related to k3s-io/k3s#2346


brandond:
Thanks for the quick response @catherineluse !

PR #2911, opened 2020-12-11
ec2:ModifyInstanceMetadataOptions resource needed

ypeek:
ec2:ModifyInstanceMetadataOptions resource needed to create EC2 instance in AWS


superseb (changes requested):
@ypeek Thanks for the PR. Can you share the error you encountered so this can be traced back to that. I don't see this requirement in the upstream policy either (possibly you found a reference somewhere, if so, please share the link)

PR #2912, opened 2020-12-12
[k3s] Update system-upgrade-controller to v0.6.2

koukihai:
Hi, I have spent longer than I would like to admit on configuring tolerations for plans, only to find out it was fixed in newer versions (>= v0.5.0) 😄 

I hope this update will save some time to other users who encountered the same issue (https://github.com/rancher/system-upgrade-controller/issues/76)!


davidnuzik:
Apologies for missing this. Thank you for the PR!

davidnuzik (approved):

PR #2914, opened 2020-12-14
Include description for security groups and private API

thedadams:
The advanced user can use security groups to enable private-only API access, but still allow Rancher to communicate with the cluster. A note about this configuration is added with a link to the AWS documentation.

Related issue:
https://github.com/rancher/rancher/issues/30403

### DIFF
--- a/content/rancher/v2.x/en/cluster-provisioning/hosted-kubernetes-clusters/eks/_index.md
+++ b/content/rancher/v2.x/en/cluster-provisioning/hosted-kubernetes-clusters/eks/_index.md
@@ -137,9 +137,11 @@ Configuring Public/Private API access is an advanced use case. For details, refe
 
 ### Private-only API Endpoints
 
-If you enable private and disable public API endpoint access when creating a cluster, then there is an extra step you must take in order for Rancher to connect to the cluster successfully. In this case, a pop-up will be displayed with instructions on how to complete the setup. Once the cluster is provisioned, you can run the displayed command anywhere you can connect to the cluster's Kubernetes API.
+If you enable private and disable public API endpoint access when creating a cluster, then there is an extra step you must take in order for Rancher to connect to the cluster successfully. In this case, a pop-up will be displayed with a command that you will run on the cluster to register it with Rancher. Once the cluster is provisioned, you can run the displayed command anywhere you can connect to the cluster's Kubernetes API.
 
-Note: you can avoid this extra step by enabling both private and public API endpoint access on cluster creation. You can disable public access after the cluster is created and in an active state and Rancher will continue to communicate with the EKS cluster.
+There are two ways to avoid this extra manual step:
+* You can create the cluster with both private and public API endpoint access on cluster creation. You can disable public access after the cluster is created and in an active state and Rancher will continue to communicate with the EKS cluster.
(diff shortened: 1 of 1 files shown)


content/rancher/v2.x/en/cluster-provisioning/hosted-kubernetes-clusters/eks/_index.md
```diff
@@ -141,6 +141,8 @@ If you enable private and disable public API endpoint access when creating a clu
 
 Note: you can avoid this extra step by enabling both private and public API endpoint access on cluster creation. You can disable public access after the cluster is created and in an active state and Rancher will continue to communicate with the EKS cluster.
 
+Additionally, if Rancher shares a subnet with the EKS cluster, then security groups can be used to enable Rancher to communicate with the cluster's API endpoint. In this case, the import command can be ignored and Rancher will be able to communicate with your cluster. For more information on configuring security groups, refer to the [security groups documentation](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_SecurityGroups.html).
```
catherineluse, line 144: I noticed that the new paragraph refers to the import command, but we didn't mention what the import command is.

I think we should explain the command, and also use the word 'registered' because we changed the name of the cluster type from imported to registered.

> If you enable private and disable public API endpoint access when creating a cluster, then there is an extra step you must take in order for Rancher to connect to the cluster successfully. In this case, a pop-up will be displayed with a command that you will run on the cluster to register it with Rancher. Once the cluster is provisioned, you can run the displayed command anywhere you can connect to the cluster's Kubernetes API.
>
> [...]
>
> Additionally, if Rancher shares a subnet with the EKS cluster, then security groups can be used to enable Rancher to communicate with the cluster's API endpoint. In this case, the command to register the cluster is not needed, and Rancher will be able to communicate with your cluster. For more information on configuring security groups, refer to the [security groups documentation](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_SecurityGroups.html).

catherineluse (changes requested):

content/rancher/v2.x/en/cluster-provisioning/hosted-kubernetes-clusters/eks/_index.md
```diff
@@ -137,10 +137,12 @@ Configuring Public/Private API access is an advanced use case. For details, refe
...
+If you enable private and disable public API endpoint access when creating a cluster, then there is an extra step you must take in order for Rancher to connect to the cluster successfully. In this case, a pop-up will be displayed with a command that you will run on the cluster to register it with Rancher. Once the cluster is provisioned, you can run the displayed command anywhere you can connect to the cluster's Kubernetes API.
 
 Note: you can avoid this extra step by enabling both private and public API endpoint access on cluster creation. You can disable public access after the cluster is created and in an active state and Rancher will continue to communicate with the EKS cluster.
 
+Additionally, if Rancher shares a subnet with the EKS cluster, then security groups can be used to enable Rancher to communicate with the cluster's API endpoint. In this case, the command to register the cluster is not needed, and Rancher will be able to communicate with your cluster. For more information on configuring security groups, refer to the [security groups documentation](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_SecurityGroups.html).
```
rmweir, line 144: I think fitting this into the above paragraph may be a good call. My concern is because of how it is worded a reader would read that step must be done and stop there to perform the step instead of noticing this paragraph. If someone disagree with this, disregard.

rmweir (changes requested):

rmweir (approved):
LGTM

PR #2054, opened 2019-12-05
Align installation requirements with terms of service

catherineluse:
Our installation docs had some inconsistencies with the terms of service. This PR updates the installation docs to match the terms of service and addresses this issue https://github.com/rancher/docs/issues/2050

PR #2925, opened 2020-12-16
RKE install add supported Docker versions

reylejano:
RKE does not support the latest Docker (at the moment of this PR creation its v20.10.0).
The following error is received with Docker v20.10.0 when running `rke up --config rancher-cluster.yml`:

> FAT[0000] Unsupported Docker version found [20.10.0] on host [IP]. supported versions are [1.13.x 17.03.x 17.06.x 17.09.x 18.06.x 18.09.x 19.03.x

This PR updates the [RKE install page](https://rancher.com/docs/rke/latest/en/installation/) with the supported Docker versions.

### DIFF
--- a/content/rke/latest/en/installation/_index.md
+++ b/content/rke/latest/en/installation/_index.md
@@ -93,7 +93,7 @@ $ port upgrade rke
 
 The Kubernetes cluster components are launched using Docker on a Linux distro. You can use any Linux you want, as long as you can install Docker on it.
 
-> RKE supports Docker versions 1.13.x, 17.03.x, 17.06.x, 17.09.x, 18.06.x, 18.09.x, 19.03.x., the latest Docker version 20.x.x is not supported.
+> For information on which Docker versions were tested with your version of RKE, refer to the [terms of service](https://rancher.com/support-maintenance-terms) for installing Rancher on RKE.
 
 Review the [OS requirements]({{<baseurl>}}/rke/latest/en/installation/os/) and configure each node appropriately.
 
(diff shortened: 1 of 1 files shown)


reylejano:
Fixes [rancher/docs issue 2926](https://github.com/rancher/docs/issues/2926)

superseb:
This was already removed for Rancher in https://github.com/rancher/docs/pull/2054, this should probably be fixed in the same way. Also, Docker installation is covered on https://github.com/rancher/docs/blob/master/content/rke/latest/en/os/_index.md#installing-docker which then should be removed similar to linked PR.

ebauman:
Adding the docker version here may be helpful because I would anticipate users who are familiar with Docker would skip over our documentation on "Installing Docker". They could then easily end up in an misconfigured state, assuming their knowledge of docker installation was sufficient. Being more explicit about the required versions would help mitigate this.

catherineluse:
As @superseb mentioned, we are moving away from saying which OS and software requirements are supported in the documentation because it easily becomes out of sync with the terms of service. 

Is it ok to say,

> For the supported Docker version, refer to the Docker requirements that correspond to your Rancher version in the [terms of service.](https://rancher.com/support-maintenance-terms)

reylejano:
> As @superseb mentioned, we are moving away from saying which OS and software requirements are supported in the documentation because it easily becomes out of sync with the terms of service.
> 
> Is it ok to say,
> 
> > For information on which Docker versions were tested with your version of RKE, refer to the [terms of service](https://rancher.com/support-maintenance-terms) for installing Rancher on RKE.

@catherineluse Yes, a note about where to find the Docker versions tested will suffice.

PR #2944, opened 2021-01-04
Improved command to get namespaces

dnoland1:
Previous command had flags like `-A` which make no sense when getting namespaces. Removed dependence on `jq` binary.


catherineluse:
@paraglade it's ok to merge this?

bl-ue (approved):

Issue #2935, opened 2020-12-23
RancherD Doc Changes


rancher-max:
**Summary:**

Reviewed RancherD docs on staging branch and have a few important edits that should be made

**Details:**

Many of these are suggestions and open for discussion, but 2 and 4 are required.

1. https://staging.rancher.com/docs/rancher/v2.x/en/installation/install-rancher-on-linux/#node-requirements
- If we link node requirements, we should probably specifically mention Docker is NOT necessary to have on the Linux Host. It currently looks like it is required when we link that page.

2. https://staging.rancher.com/docs/rancher/v2.x/en/installation/install-rancher-on-linux/#1-set-up-the-config-yaml in "If you do not specify a pre-shared secret, RancherD will generate one and place it at /var/lib/rancher/rancherd/server/node-token."
- this should be  /var/lib/rancher/rke2/server/node-token

3. https://staging.rancher.com/docs/rancher/v2.x/en/installation/install-rancher-on-linux/#2-launch-the-first-server-node
- In step 1 here, should we specify available environment variables? The relevant one to me is: `INSTALL_RANCHERD_VERSION`
- So an example curl command would be like: `curl -sfL https://get.rancher.io | INSTALL_RANCHERD_VERSION=v2.5.4-rc6 sh -`

4. https://staging.rancher.com/docs/rancher/v2.x/en/installation/install-rancher-on-linux/rancherd-configuration/#customizing-the-rancherd-helm-chart In "Put this manifest on your host in /var/lib/rancher/rancherd/server/manifests before running RancherD."
- should be /var/lib/rancher/rke2/server/manifests

5. https://staging.rancher.com/docs/rancher/v2.x/en/installation/install-rancher-on-linux/#4-verify-that-rancher-is-installed-on-the-kubernetes-cluster 
Should we mention that there will be many things installed? If watching pods, you'll see:
  - `helm-operation` pods in `cattle-system` namespace
  - a `rancher` pod and `rancher-webhook` pod in `cattle-system` namespace
  - a `fleet-agent`, `fleet-controller`, and `gitjob` pod in `fleet-system` namespace
  - a `rancher-operator` pod in `rancher-operator-system` namespace
 
6. https://staging.rancher.com/docs/rancher/v2.x/en/installation/install-rancher-on-linux/#fixed-registration-address
[...]

PR #2947, opened 2021-01-05
Address feedback on RancherD docs #2935

catherineluse:
Closes https://github.com/rancher/docs/issues/2935

@rancher-max Should we call out any more environment variables than the rancherd version one? That's the only rancherd environment variable I know of.


rancher-max:
> @rancher-max Should we call out any more environment variables than the rancherd version one? That's the only rancherd environment variable I know of.

That is the most relevant one, and I like it the way you have it so I would say let's only include that. 

For reference, the other environment variables come from https://github.com/rancher/rancher/blob/master/cmd/rancherd/install.sh and the available options for channel are from https://update.rancher.io/v1-release/channels. I assume at some point we will have a `stable` channel as well that may become default, and at that point it might be useful to document this variable.

rancher-max (approved):
LGTM, these changes are perfect! Thank you!

Issue #2948, opened 2021-01-05
[Rancher2]  GKE auto-upgrade recommendation


cbron:
Related rancher issue: https://github.com/rancher/rancher/issues/29988
In short, don't recommend auto-upgrading GKE nodes
While some code is going into the new releases, this change is not release specific.

PR #2949, opened 2021-01-05
Recommend not auto-upgrading GKE nodes

cbron:
https://github.com/rancher/docs/issues/2948


cbron:
We could leave it all as one sentence, but I thought this might be cleaner way to add a note to specific part.

catherineluse (approved):
Lgtm

PR #2957, opened 2021-01-07
Added Oxford comma

Tejeev:
In my experience, it can go a long way to clarifying sentences and we use it all over the docs (including many times on this very page) so added it here


catherineluse:
Some style guides say the Oxford comma should only be used when it does prevent ambiguity.

But the Microsoft Manual of Style, which is what I try to use most of the time, says yes to the Oxford comma. So I'm with you on this one.

PR #3009, opened 2021-02-08
Added doc note FW rule for GKE private clusters

bluemalkin:
Hello,

Raising a PR for an important note on how to get Rancher working correctly in a GKE private cluster.

Fixes https://github.com/rancher/rancher/issues/29451


catherineluse:
lgtm

PR #3012, opened 2021-02-09
Update _index.md

rssfed23:
Updated CNI under values that was missing an indentation and throwing a syntax error in the Cluster Explorer. As per https://istio.io/latest/docs/setup/additional-setup/cni/#helm-chart-parameters, it should be indented.
Credit to @it-slav for identifying the bug and fix


rssfed23:
Closing as this was created in error and @it-slav has since successfully installed Istio with the "invalid" yaml

Issue #3003, opened 2021-02-05
[Rancher2] [Fleet] Add doc for Windows clusters fleet agent on Rancher pre-2.5.6


izaac:
**Summary:**
On Windows clusters pre-2.5.6 the fleet-agent is observed getting in Error state 

This is a known limitation but we need instructions for users that have their fleet agent in Error to deploy the agent in the proper Linux node.

This is specially important because if a user expects to upgrade the fleet-agent it first has to be in Active state and registered with the fleet-controller.

**Details:**

### We need to make sure first this are the Rancher recommended steps to follow

This is how I was able to provision the fleet-agent workload in the Linux worker.

- Cordon the Windows Nodes
- Apply the below toleration to the fleet-agent workload
- Re-deploy the fleet-agent

```yaml
tolerations:
- effect: NoSchedule
  key: cattle.io/os
  operator: Equal
  value: linux
```

Rancher 2.5.6 Windows support will be introduced so we don't expect this to happen for new clusters.
Reference issue: https://github.com/rancher/rancher/issues/30516

PR #3014, opened 2021-02-09
[Rancher2] Add 2.5.6 Fleet upgrade workflow for Windows

nickgerace:
Relevant issue: https://github.com/rancher/docs/issues/3003

### DIFF
--- a/content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
+++ b/content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
@@ -21,7 +21,8 @@ Fleet comes preinstalled in Rancher v2.5. To access it, go to the **Cluster Expl
 
 ### Windows Support
 
-Prior to Rancher v2.5.6, the `agent` would fail on downstream clusters with Windows nodes.
+Prior to Rancher v2.5.6, the `agent` did not have native Windows manifests on downstream clusters with Windows nodes.
+This would result in a failing `agent` pod for the cluster.
 If you are upgrading from an older version of Rancher to v2.5.6+, you can deploy a working `agent` with the following workflow *in the downstream cluster*:
 
 1. Cordon all Windows nodes.
--- a/content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
+++ b/content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
@@ -19,6 +19,25 @@ deploy everything in the cluster. This give a high degree of control, consistenc
 
 Fleet comes preinstalled in Rancher v2.5. To access it, go to the **Cluster Explorer** in the Rancher UI. In the top left dropdown menu, click **Cluster Explorer > Continuous Delivery.** On this page, you can edit Kubernetes resources and cluster groups managed by Fleet.
 
+### Windows Support
+
+Prior to Rancher v2.5.6, the `agent` would fail on downstream clusters with Windows nodes.
+If you are upgrading from an older version of Rancher to v2.5.6+, you can deploy a working `agent` with the following workflow *in the downstream cluster*:
+
+1. Cordon all Windows nodes.
+1. Apply the below toleration to the `agent` workload.
+1. Uncordon all Windows nodes.
+1. Delete all `agent` pods. New pods should be created with the new toleration.
+1. Once the `agent` pods are running, and auto-update is enabled for Fleet, they should be updated to a Windows-compatible `agent` version.
+
+```yaml
+tolerations:
+- effect: NoSchedule
+  key: cattle.io/os
+  operator: Equal
+  value: linux
+```
+
 ### GitHub Repository
 
 The Fleet Helm charts are available [here.](https://github.com/rancher/fleet/releases/latest)


content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
```diff
@@ -19,10 +19,29 @@ deploy everything in the cluster. This give a high degree of control, consistenc
 
 Fleet comes preinstalled in Rancher v2.5. To access it, go to the **Cluster Explorer** in the Rancher UI. In the top left dropdown menu, click **Cluster Explorer > Continuous Delivery.** On this page, you can edit Kubernetes resources and cluster groups managed by Fleet.
 
+### Windows Support
+
+Prior to Rancher v2.5.6, the `agent` would fail on downstream clusters with Windows ndoes.
```
StrongMonkey, line 24:
```suggestion
Prior to Rancher v2.5.6, the `agent` would fail on downstream clusters with Windows nodes.
```

StrongMonkey:
LGTM otherwise :100:

content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
nickgerace: oops. Thank you!

content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
```diff
@@ -19,10 +19,29 @@ deploy everything in the cluster. This give a high degree of control, consistenc
 
 Fleet comes preinstalled in Rancher v2.5. To access it, go to the **Cluster Explorer** in the Rancher UI. In the top left dropdown menu, click **Cluster Explorer > Continuous Delivery.** On this page, you can edit Kubernetes resources and cluster groups managed by Fleet.
 
+### Windows Support
+
+Prior to Rancher v2.5.6, the `agent` would fail on downstream clusters with Windows nodes.
+If you are upgrading from an older version of Rancher to v2.5.6+, you can deploy a working `agent` with the following workflow *in the downstream cluster*:
```
thedadams, line 25: The bolding above has two asterisks. I would recommend doing the same here to ensure it shows up correctly in the docs.

content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
```diff
@@ -19,10 +19,29 @@ deploy everything in the cluster. This give a high degree of control, consistenc
...
+
+1. Cordon all Windows nodes.
```
thedadams, line 27: Are we sure that the markdown will be converted to a numbered list correctly? If yes, cool. If not, I would change these to be numbered.

content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
nickgerace: This is meant to be italics. Perhaps, we should exchange italics for bolding?

content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
nickgerace: Yes! I viewed it using `scripts/dev` and originally found the pattern in other section of the docs.

thedadams (approved):

content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
```diff
@@ -19,10 +19,29 @@ deploy everything in the cluster. This give a high degree of control, consistenc
 
 Fleet comes preinstalled in Rancher v2.5. To access it, go to the **Cluster Explorer** in the Rancher UI. In the top left dropdown menu, click **Cluster Explorer > Continuous Delivery.** On this page, you can edit Kubernetes resources and cluster groups managed by Fleet.
 
+### Windows Support
+
+Prior to Rancher v2.5.6, the `agent` would fail on downstream clusters with Windows nodes.
```
cbron, line 24: Better phrasing would be that the agent did not have windows images/manifests, rather than it would just fail.

content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
cbron, line 27: Why do we have to cordon the nodes ? Why not just add toleration and delete the pod.

content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
nickgerace: Fixing. Good point

content/rancher/v2.x/en/deploy-across-clusters/fleet/_index.md
nickgerace: The manifest will not be changed to a Linux one from our testing. The deleted pod was re-created using the same manifest. I can add an `(optional)` tag next to the step to indicate that your mileage may vary. In our reproduction, we had to cordon the node, but this could have been timing, human error, etc.

cbron: Weird, @StrongMonkey or @luthermonson  you know why that would be the case ? I don't think we should put optional, ideally we figure out correct way.

StrongMonkey: Oh, I think we can just cordon the windows nodes and delete the pod. Don't need to add tolerations since it will be added after upgrade.

nickgerace: I think @cbron's concern is cordoning windows nodes that may be running other pods with Windows containers unrelated to Fleet.

nickgerace: Wait @cbron I think I am accidentally talking past you here. We `cordon` because we want the node to be _unschedule-able_. We don't `drain` the node, so I do not think cordoning is an issue here, actually. For some reason, I was thinking `cordon == drain` in my first reply to your comment. Given this context, I believe this step should stay in the instructions.

cbron: Ahh yes I was also thinking drain. Just cordon makes sense, so the deleted pod doesn't get scheduled on windows.

PR #3030, opened 2021-02-15
Remove dependency on jq in example

dkeightley:


catherineluse:
@paraglade is this ok to merge?

macedogm:
@dkeightley Would you like to update this for the v2.6 hardening guides too - https://github.com/rancher/docs/pull/3810 ?

PR #3086, opened 2021-03-03
add alert to docs

jrcreative:


jrcreative:
@catherineluse here is the PR with the alert and the update to the redirect

PR #3103, opened 2021-03-12
Adjust Chart Directory Structure

alexarnoldy:
Adjusted the diagram so it looks more like a tree output and better shows the directory structure. 

Adjusted the description above it, since none of the items in the diagram say "Rancher Specific".

### DIFF
--- a/content/rancher/v2.0-v2.4/en/helm-charts/creating-apps/_index.md
+++ b/content/rancher/v2.0-v2.4/en/helm-charts/creating-apps/_index.md
@@ -45,18 +45,20 @@ Advantages of Rancher charts include:
 
 # Chart Directory Structure
 
-The following table demonstrates the directory structure for a chart, which can be found in a chart directory: `charts/<APPLICATION>/<APP_VERSION>/`. This information is helpful when customizing charts for a custom catalog. Files denoted with **Rancher Specific** are specific to Rancher charts, but are optional for chart customization.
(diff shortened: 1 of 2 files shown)


catherineluse:
Hi Alex. Can you please make these changes in the v2.0-v2.4 directory?  (content/rancher/v2.0-v2.4) Because the content in `content/rancher/v2.x` is no longer being maintained.

alexarnoldy:
Wow, I totally had that wrong. I thought 2.x was the path the covers everything in common with the various versions. Would it be more appropriate to update the 2.5 material?

Best regards,

--Alex.


Alex Arnoldy

Software Defined Infrastructure Solution Architect

[cid:60920256-571c-45d3-86dd-c3d5a90be587]

SUSE: Simplify, Modernize, Accelerate

SUSE One Partners

The Power of Many. Together as One.

mobile: 408.828.9797

email: ***@***.******@***.***>

________________________________

catherineluse:
Yes, please update the v2.5 material.

We started out with 2.x. Then that became confusing and hard to read because we had all kinds of information for different versions mixed together. So was hard to see relevant info because of the large amount if irrelevant info.

To make the 2.5 directory I stripped out all the versions relevant to v2.0-2.4.

And to make the earlier docs easier to read as well, I also made a 2.0-.2.4 directory with all the 2.5 content stripped out.

Unfortunately we can't remove the 2.x section because then people would click on Google search results and get 404 errors.

In future I hear we're going to move to a better system where the docs for each product will be in a separate repo and there will be separate git branches for each version.

btat:
I couldn't find the equivalent content in the v2.5 directory, so only applied the changes to the v2.0-2.4 directory.

PR #3111, opened 2021-03-16
in tree is depricated.  see: https://github.com/kubernetes/enhancemen…

Tejeev:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-cloud-provider/2395-removing-in-tree-cloud-providers#phase-4---disabling-in-tree-providers


deniseschannon (commented):
We should not add this info until we have a proper solution of how we will support external/out of tree cloud providers.

PR #3118, opened 2021-03-17
Make clear ipsec flannel backend is not functional

txomon:
Currently ipsec backend is a recommended set up method that doesn't work. Until https://github.com/k3s-io/k3s/issues/1613 is fixed, mark the ipsec backend as not functional


davidnuzik:
Going to close this as per https://github.com/k3s-io/k3s/issues/1613#issuecomment-865857158

PR #3119, opened 2021-03-17
Change docker.io to mycustomreg

dharmendrakariya:


brandond:
Not sure what the point of this is - the documentation shows an example of using `https://mycustomreg.com:5000` as a local registry mirror for docker.io (docker hub).

Issue #3147, opened 2021-03-29
Document project network isolation on any cluster + CNI


catherineluse:
https://rancher.zendesk.com/agent/tickets/15139
Point of contact: Ryan

PR: https://github.com/rancher/rancher/pull/31344

Jan said:

> If the RKE network plugin supports the enforcement of Kubernetes network policies (as it the case for Cisco ACI plugin) then Rancher should allow the user to enable project network isolation as a cluster option.
> This Rancher feature integrates with the Kubernetes network policy subsystem and not with a specific CNI. The validation of that option should therefore depend on whether the CNI supports enforcement of the Rancher managed policies and not whether that CNI is Canal.

https://github.com/rancher/rancher/issues/31338
zendesk ticket 15139

catherineluse:
We have documented that when the project network isolation option is enabled with Canal, the system project is exempt from the isolation. Would the same be true for any CNI that supports enforcement of Kubernetes network policies?

>**Note:** In clusters where both:
>
> - The Canal network plug-in is in use.
> - The Project Network Isolation option is enabled.
>
>The `system` project overrides the Project Network Isolation option so that it can communicate with other projects, collect logs, and check health.

PR #3175, opened 2021-04-11
Document project network isolation on any cluster + CNI

catherineluse:
https://github.com/rancher/docs/issues/3147

We have documented that when the project network isolation option is enabled with Canal, the system project is exempt from the isolation. Would the same be true for any CNI that supports enforcement of Kubernetes network policies?

>**Note:** In clusters where both:
>
> - The Canal network plug-in is in use.
> - The Project Network Isolation option is enabled.
>
>The `system` project overrides the Project Network Isolation option so that it can communicate with other projects, collect logs, and check health.


ryansann (approved):

ryansann:
@catherineluse that is still true about the system project, it is able to communicate with all projects and all projects accept traffic from it

catherineluse:
The Istio docs mentioned extra steps for clusters with project network isolation. I've edited them assuming that those extra steps also apply to v2.5.8 clusters in which a CNI plugin other than Canal is being used. @brendarearden is it accurate that those extra steps still apply when something other than Canal is used?

deniseschannon (approved):

Issue #3105, opened 2021-03-12
[Rancher2]  Update GKE docs for GKEv2


cbron:
Specifics Include:
* logs viewer permissions is needed
* How to use setting interval

Catherine edit April 23rd:
The same roles from the GKEv1 docs worked for GKEv2 as well, so the documented permissions won't change.

catherineluse:
Point of contact: Colleen

gaktive:
Once docs are in place, this will help with additional UI pointers: https://github.com/rancher/rancher/issues/32030

catherineluse:
@btat I've added partial docs in staging to update the information about GKE cluster configuration.

The changes are in this PR https://github.com/rancher/docs/pull/3200 

- The new configuration reference page is at: https://staging.rancher.com/docs/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/config-reference/ 
  - On this page I added a tab for v2.5.8 that explains the new cluster provisioning. It now says you need a project ID and it breaks out creating cloud credentials as a separate step.
- And the updated GKE cluster provisioning page is at: https://staging.rancher.com/docs/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/ 
  - On this page I created a new GKE cluster configuration reference because it didn't already exist for Rancher before v2.5.8. So I created both tabs. I filled in the second half of the fields.   

Can you please create a new PR that fills in the incomplete information in the configuration reference? I've added placeholder section titles for the fields. (Or at least, I think I got all the fields. Let me know if you notice anything missing.)

There will also need to be changes to the docs regarding registered GKE clusters, but that's a different page and can be handled with a separate PR.

I've organized the configuration reference based on screenshots from GKE cluster provisioning before and after v2.5.8. This is how I got the organization to break down the reference options by cluster options, node pools etc.

# What still needs to be documented

- First half of configuration reference (the ones that I added placeholders for)
- Logs viewer permissions is needed
- How to use setting interval
- Registered cluster docs for GKE

# Screenshots of GKE provisioning in v2.5.8
<img width="1311" alt="Screen Shot 2021-04-14 at 8 38 43 AM" src="https://user-images.githubusercontent.com/20599230/115448903-16567880-a1cf-11eb-8949-fe6b54ef0086.png">
[...]

PR #3188, opened 2021-04-14
[KEv2] Add note on GKE private clusters

cmurphy:
The new KEv2-driven GKE cluster provider will support creating "private
clusters", but this has a unique meaning for GKE that is different from
other Kubernetes providers that needs some extra explanation, and can
require pre-planning and/or manual steps during the cluster provisioning
process. This change documents the considerations needed for an advanced
setup like this.

Partial https://github.com/rancher/docs/issues/3105

### DIFF
--- a/content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
+++ b/content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
@@ -73,6 +73,46 @@ You can access your cluster after its state is updated to **Active.**
 - `Default`, containing the `default` namespace
 - `System`, containing the `cattle-system`, `ingress-nginx`, `kube-public`, and `kube-system` namespaces
 
+## Private Clusters
+
+In GKE, [private clusters](https://cloud.google.com/kubernetes-engine/docs/concepts/private-cluster-concept) are clusters whose nodes are isolated from inbound and outbound traffic by assigning them internal IP addresses only. Private clusters in GKE have the option of exposing the control plane endpoint as a publicly accessible address or as a private address. This is different from other Kubernetes providers, which may refer to clusters with private control plane endpoints as "private clusters" but still allow traffic to and from nodes. You may want to create a cluster with private nodes, with or without a public control plane endpoint, depending on your organization's networking and security requirements. A GKE cluster hosted in Rancher can use isolated nodes by selecting "Private Cluster" in the Cluster Options (under "Show advanced options"). The control plane endpoint can optionally be made private by selecting "Enable Private Endpoint".
+
+### Private Nodes
+
+Because the nodes in a private cluster only have internal IP addresses, they will not be able to install the cluster agent and Rancher will not be able to fully manage the cluster. This can be overcome in a few ways.
+
+#### Cloud NAT
+
+>**Note**
+>Cloud NAT will [incur charges](https://cloud.google.com/nat/pricing).
+
+If restricting outgoing internet access is not a concern for your organization, use Google's [Cloud NAT](https://cloud.google.com/nat/docs/using-nat) service to allow nodes in the private network to access the internet, allowing them to download the required images from Dockerhub. This is the simplest solution.
+
+#### Private registry
+
+>**Note**
+>This scenario is not officially supported, but is described for cases in which using the Cloud NAT service is not sufficient.
+
+If restricting both incoming and outgoing traffic to nodes is a requirement, follow the air-gapped installation instructions to set up a private container image [registry](https://rancher.com/docs/rancher/v2.x/en/installation/other-installation-methods/air-gap/) on the VPC where the cluster is going to be, allowing the cluster nodes to access and download the images they need to run the cluster agent.
+
+### Private Control Plane Endpoint
+
+If the cluster has a public endpoint exposed, Rancher will be able to reach the cluster, and no additional steps need to be taken. However, if the cluster has no public endpoint, then considerations must be made to ensure Rancher can access the cluster.
+
+#### Cloud NAT
+
+>**Note**
+>Cloud NAT will [incur charges](https://cloud.google.com/nat/pricing).
+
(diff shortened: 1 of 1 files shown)


catherineluse:
These changes look good to me. Can you please rebase on the staging branch?

cmurphy:
@catherineluse I've rebased and changed the base branch to staging, can you take another look?

cmurphy:
Marked the private registry scenario as not officially supported, based on discussion with QA

catherineluse (approved):

content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
```diff
@@ -73,6 +73,46 @@ You can access your cluster after its state is updated to **Active.**
 - `Default`, containing the `default` namespace
 - `System`, containing the `cattle-system`, `ingress-nginx`, `kube-public`, and `kube-system` namespaces
 
+## Private Clusters
+
+In GKE, [private clusters](https://cloud.google.com/kubernetes-engine/docs/concepts/private-cluster-concept) are clusters whose nodes are isolated from inbound and outbound traffic by assigning them internal IP addresses only. Private clusters in GKE have the option of exposing the control plane endpoint as a publicly accessible address or as a private address. This is different from other Kubernetes providers, which may refer to clusters with private control plane endpoints as "private clusters" but still allow traffic to and from nodes. You may want to create a cluster with private nodes, with or without a public control plane endpoint, depending on your organization's networking and security requirements. A GKE cluster hosted in Rancher can use isolated nodes by selecting "Private Cluster" in the Cluster Options (under "Show advanced options"). The control plane endpoint can optionally be made private by selecting "Enable Private Endpoint".
```
cbron, line 78: > A GKE cluster hosted in Rancher 

A GKE cluster provisioned from Rancher

content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
```diff
@@ -73,6 +73,46 @@ You can access your cluster after its state is updated to **Active.**
...
+### Private Nodes
+
+Because the nodes in a private cluster only have internal IP addresses, they will not be able to install the cluster agent and Rancher will not be able to fully manage the cluster. This can be overcome in a few ways.
+
+#### Cloud NAT
+
+>**Note**
+>Cloud NAT will [incur charges](https://cloud.google.com/nat/pricing).
+
+If restricting outgoing internet access is not a concern for your organization, use Google's [Cloud NAT](https://cloud.google.com/nat/docs/using-nat) service to allow nodes in the private network to access the internet, allowing them to download the required images from Dockerhub. This is the simplest solution.
```
cbron, line 89: > allowing them to download the required images from Dockerhub

and contact the Rancher management server.

content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
```diff
@@ -73,6 +73,46 @@ You can access your cluster after its state is updated to **Active.**
...
+
+#### Private registry
+
+>**Note**
+>This scenario is not officially supported, but is described for cases in which using the Cloud NAT service is not sufficient.
+
+If restricting both incoming and outgoing traffic to nodes is a requirement, follow the air-gapped installation instructions to set up a private container image [registry](https://rancher.com/docs/rancher/v2.x/en/installation/other-installation-methods/air-gap/) on the VPC where the cluster is going to be, allowing the cluster nodes to access and download the images they need to run the cluster agent.
```
cbron, line 96: This can only work if there is a public endpoint right ? Otherwise it can't if there is no outbound connection from the nodes. We should either specify that or remove this section.

content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
```diff
@@ -73,6 +73,46 @@ You can access your cluster after its state is updated to **Active.**
...
+
+### Private Control Plane Endpoint
```
cbron, line 98: I think this section should go above the Node part. The user must understand that they need a public endpoint, open NAT, or direct access first, before creating the nodes.

content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
cmurphy: No, if there's only a private endpoint this should still work as long as Rancher is running on the same VPC or is connected through a VPN.

content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
cmurphy: I chose to place it below because the private control plane endpoint is only an option if you also have private nodes. So the private nodes section above is the most generally applicable, and this section is an optional choice after that. It's also how it shows up in the Rancher UI and in the GKE console.

content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
thedadams: If the nodes are denied outbound traffic and Rancher is not running on the same VPC as the cluster then this will not allow the cluster agent to connect to the Rancher server regardless of whether the API endpoint is public. If the API endpoint is public then Rancher will still likely be able to manage the cluster, but the cluster agent will be deployed on the GKE cluster taking up resources.

There may be other consequences I am not aware of.

content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
cbron: Ok, lets add some qualifiers in section above then.

content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/_index.md
cbron: Then we need to specify that direct access by Rancher is necessary for this to work.

PR #3207, opened 2021-04-21
Update v2.5.8 GKE config reference

cmurphy:
* Add notes about Shared VPCs
* Change "Node Subnet" to "Node Subnet / Subnet" - the field name
  changes depending on whether you use Ip Aliases, but means the same
  thing in both cases
* Clarify use of Subnetwork Name
* Clarify use of address ranges and range names
* Add private cluster-related fields
* Fix HTTP (L7) Load Balancing reference URL
* Add reference URL for Maintenance Window
* Add notes on image types
  - https://github.com/rancher/rancher/issues/32211
  - https://github.com/rancher/rancher/issues/32161
* Enhance Machine Type
* Use "node pool" instead of "node group"
* Fix "Autoscaling" under "Node Pools" - this has been fixed in the UI
* Add note to "Auto Upgrade" to match <2.5.8 tab

https://github.com/rancher/docs/issues/3105


catherineluse (approved):
These changes look great. Thanks Colleen.

Issue #3133, opened 2021-03-26
[Rancher2] Update EKSv2 docs to clarify cron settings conversion


cbron:
Users should be aware of setting that changed, how rancher deals with it, and how to convert from one to the other.

PR #3210, opened 2021-04-22
Document refresh setting change in v2.5.8 #3133

catherineluse:
https://github.com/rancher/docs/issues/3133

### DIFF
--- a/content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/eks/_index.md
+++ b/content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/eks/_index.md
@@ -738,4 +738,27 @@ If Rancher and another source attempt to update an EKS cluster at the same time 
 
 ### Configuring the Refresh Interval
 
-It is possible to change the refresh interval through the setting “eks-refresh-cron". This setting accepts values in the Cron format. The default is `*/5 * * * *`. The shorter the refresh window is the less likely any race conditions will occur, but it does increase the likelihood of encountering request limits that may be in place for AWS APIs.
+{{% tabs %}}
+{{% tab "Rancher v2.5.8+" %}}
+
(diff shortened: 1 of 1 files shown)


content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/eks/_index.md
```diff
@@ -738,4 +738,27 @@ If Rancher and another source attempt to update an EKS cluster at the same time
 
 ### Configuring the Refresh Interval
 
-It is possible to change the refresh interval through the setting “eks-refresh-cron". This setting accepts values in the Cron format. The default is `*/5 * * * *`. The shorter the refresh window is the less likely any race conditions will occur, but it does increase the likelihood of encountering request limits that may be in place for AWS APIs.
+{{% tabs %}}
+{{% tab "Rancher v2.5.8+" %}}
+
+The `eks-refresh-cron` setting is deprecated. It has been migrated to the `eks-refresh` setting, which is an integer in seconds.
```
thedadams, line 744: To me, the term "deprecated" means that something is still usable, but it will be removed in a future release. Other than migrating `eks-refresh-cron` to the new setting on upgrade, it is being ignored in v2.5.8+.

I think that using "obsolete" would better represent how it is being ignored.

cbron: I'd say deprecated is ok. We'd like to remove in future release but we don't have plans to at the moment because we run into a race condition with the migration. 2.7 is earliest we could realistically remove it as we don't support upgrading more than one minor version at a time.

cbron, line 744:
```suggestion
The `eks-refresh-cron` setting is deprecated. It has been migrated to the `eks-refresh` setting, which is an integer representing seconds.
```

PR #3227, opened 2021-04-29
Add note about region and default instance type

rmweir:
**Issue:**
https://github.com/rancher/rancher/issues/31980

### DIFF
--- a/content/rancher/v2.5/en/cluster-provisioning/rke-clusters/node-pools/ec2/ec2-node-template-config/_index.md
+++ b/content/rancher/v2.5/en/cluster-provisioning/rke-clusters/node-pools/ec2/ec2-node-template-config/_index.md
@@ -36,7 +36,7 @@ Please refer to [Amazon EC2 security group when using Node Driver]({{<baseurl>}}
 
 ### Instance Options
 
-Configure the instances that will be created. Make sure you configure the correct **SSH User** for the configured AMI. It is possible that a selected region does not support the default instance type. Using the default in the aforementioned scenario will result in an error stating the requested configuration is not supported.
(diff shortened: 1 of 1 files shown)


content/rancher/v2.5/en/cluster-provisioning/rke-clusters/node-pools/ec2/ec2-node-template-config/_index.md
```diff
@@ -36,7 +36,7 @@ Please refer to [Amazon EC2 security group when using Node Driver]({{<baseurl>}}
 
 ### Instance Options
 
-Configure the instances that will be created. Make sure you configure the correct **SSH User** for the configured AMI.
+Configure the instances that will be created. Make sure you configure the correct **SSH User** for the configured AMI. It is possible that a selected region does not support the default instance type. Using the default in the aforementioned scenario will result in an error stating the requested configuration is not supported.
```
cbron, line 39: Nit: 
```suggestion
Configure the instances that will be created. Make sure you configure the correct **SSH User** for the configured AMI. It is possible that a selected region does not support the default instance type. In this scenario you must select an instance type that does exist, otherwise an error will occur stating the requested configuration is not supported.
```

cbron (approved):

content/rancher/v2.5/en/cluster-provisioning/rke-clusters/node-pools/ec2/ec2-node-template-config/_index.md
cbron: This may be fixed by UI in time for 2.5.8, but this sort of statement is ok as there may be other regions we don't know about.

Issue #3246, opened 2021-05-07
[Rancher2] Add Windows/Linux feature parity chart to Rancher 2 documentation


sirredbeard:
**Request Summary:**
I would like to add a Windows:Linux feature parity chart to the Rancher 2 documentation.

**Details:**
I have a draft PR I would to propose.

catherineluse:
Closing because https://github.com/rancher/docs/pull/3247 was merged

PR #3247, opened 2021-05-07
Create Windows Linux feature parity matrix

sirredbeard:
See issue https://github.com/rancher/docs/issues/3246


catherineluse:
I think we should tie the table to Rancher v2.5.8 instead of the date. When more windows features are added, we can create tabs on this page with separate tables for each Rancher version.

PR #3263, opened 2021-05-12
Change "Defintion" to "Definition" in the diagram

jagwire16:


catherineluse:
Closing because this should be an issue instead of a pull request

PR #3264, opened 2021-05-12
Add "Mutable" indicator for GKE config-reference

cmurphy:
The Rancher UI prevents a user from editing a field for a provisioned
GKE cluster that GKE does not support changing, but the Rancher API
schema allows updates to some fields that will get ignored and cause
confusion to API users. This change (for 2.5.8 only) tries to mitigate
that confusion by pointing out which fields can be changed.

### DIFF
--- a/content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/config-reference/_index.md
+++ b/content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/config-reference/_index.md
@@ -33,79 +33,79 @@ weight: 1
 
 ### Kubernetes Version
 
-**Mutable: yes**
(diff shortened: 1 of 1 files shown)


thedadams (approved):

content/rancher/v2.5/en/cluster-provisioning/hosted-kubernetes-clusters/gke/config-reference/_index.md
```diff
@@ -33,54 +33,80 @@ weight: 1
 
 ### Kubernetes Version
 
+**Mutable: yes**
```
cbron, line 36: nit: consider italics 😄 , just because to me the bold and header are too close.
More of a point for @catherineluse

catherineluse (approved):

PR #3278, opened 2021-05-19
Update Istio uninstall steps

### LINKED ISSUE #3023: [Rancher2] Istio V2 - Update Uninstall Steps and Notes
**Request Summary:**
Istio uninstall steps have changed. This is due to errors that were occurring on uninstall (see: https://github.com/rancher/rancher/issues/31046).

Uninstall will change from a 2 step process to a 3 step process and some information around the specific warning should be added for ease of search. 

**Details:**
https://github.com/rancher/docs/compare/master...brendarearden:istio-uninstall?expand=1
Changes for the chart readme were also made if you want to reference those: https://github.com/rancher/charts/pull/990

Add change to: https://github.com/rancher/docs/blob/master/content/rancher/v2.x/en/istio/v2.5/disabling-istio/_index.md
Change will be in the `Uninstall Istio in a Cluster` section
Change uninstall steps to: 
```
1. From the **Cluster Explorer,** navigate to **Installed Apps** in **Apps & Marketplace** and locate the `rancher-istio` installation.
1. Select `rancher-istio` in the `istio-system namespace and click **Delete**
1. After `rancher-istio` is deleted, you can then select all the remaining apps in the `istio-system` namespace and click **Delete**
```

If it makes sense to, add a troubleshooting section of some sort for the following information 

```
**Troubleshooting Uninstall:** If you didn't follow the uninstall steps, you may encounter a warning during uninstall:

`Error: uninstallation completed with 1 error(s): unable to build kubernetes objects for delete: unable to recognize "": no matches for kind "MonitoringDashboard" in version "monitoring.kiali.io/v1alpha1"`

This could mean a few things. You either selected all the apps in the `istio-system` namespace and deleted them at the same time, or you deleted `rancher-istio` chart dependencies prior to deleting the `rancher-istio` chart. Since the uninstall did not complete properly, you will have resources remaining in the `istio-system` namespace that you will need to manually clean up. Another option to avoid manual clean-up is to install `rancher-istio` again, then uninstall it in the correct order.
```

btat:
Closes rancher/docs#3023

### DIFF
--- a/content/rancher/v2.5/en/istio/disabling-istio/_index.md
+++ b/content/rancher/v2.5/en/istio/disabling-istio/_index.md
@@ -12,11 +12,18 @@ This section describes how to uninstall Istio in a cluster or disable a namespac
 To uninstall Istio,
 
 1. From the **Cluster Explorer,** navigate to **Installed Apps** in **Apps & Marketplace** and locate the `rancher-istio` installation.
-1. Select all the apps in the `istio-system` namespace and click  **Delete**. 
+1. Select `rancher-istio` in the `istio-system namespace and click **Delete**
+1. After `rancher-istio` is deleted, you can then select all the remaining apps in the `istio-system` namespace and click **Delete**
 
 **Result:** The `rancher-istio` app in the cluster gets removed. The Istio sidecar cannot be deployed on any workloads in the cluster. 
 
-**Note:** You can no longer disable and reenable your Istio installation. If you would like to save your settings for a future install, view and save individual YAMLs to refer back to / reuse for future installations. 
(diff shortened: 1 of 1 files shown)


btat:
> 
> 
> What version of Rancher did this change in? v2.5.8? If so we should have a 2.5.8 tab and an older version tab to add the new info while preserving the older docs.

In https://github.com/rancher/rancher/issues/31046 the bug was reported for Istio 1.7.x and looking at release notes Istio 1.7 was added to Rancher v2.5, so the fix should apply to all v2.5 docs. If I understand correctly, updating it for all of the v2.5.x docs would be preferred since it's a minor inconvenience for users at most because it's only splitting one command into two, but it avoids the potential of the bug in 31046 occurring.

Issue #3234, opened 2021-05-01
[Rancher2] Monitoring-alerting upgrade - use of RKE template can re-enable legacy monitoring


krumware:
**Request Summary:**

in: content/rancher/v2.5/en/monitoring-alerting/migrating/_index.md
https://rancher.com/docs/rancher/v2.5/en/monitoring-alerting/migrating/

While following the upgrade guide, I could not understand why legacy monitoring was being re-enabled while updating my clusters that were managed with RKE templates (to update kubernetes version), and did not realize that monitoring was enabled within the template. Each time the cluster was modified, the legacy monitoring would be re-deployed and the upgrade steps would need to be repeated.  I do not know what would happen if I had enabled new monitoring then made RKE changes, without first disabling monitoring in the RKE template.

**Details:**

To: https://github.com/rancher/docs/blob/master/content/rancher/v2.5/en/monitoring-alerting/migrating/_index.md#migrating-from-monitoring-v1-to-monitoring-v2 I would like to add: `If the cluster is managed using an RKE template: To prevent legacy monitoring from being re-enabled, disable Monitoring in future RKE template revisions.`

https://github.com/krumware/docs-1/blob/patch-1/content/rancher/v2.5/en/monitoring-alerting/migrating/_index.md#migrating-from-monitoring-v1-to-monitoring-v2

krumware:
referencing: https://github.com/rancher/rancher/issues/32416

referencing: https://github.com/bashofmann/rancher-monitoring-v1-to-v2#check-why-monitoring-v1-is-not-disabled

Some reinforcement here, 
Running the check_monitoring_disabled helper on clusters with this affected will return 
```
The RKE template revision used to spin up the cluster enables Monitoring / Alerting V1. You cannot switch to Monitoring V2 until you have updated your RKE template to set enable_cluster_alerting=False and enable_cluster_monitoring=False.
```

PR #3282, opened 2021-05-19
Add required modifications to RKE templates for monitoring migration

krumware:
closes: https://github.com/rancher/docs/issues/3234

The migration documentation does not cover cases where RKE templates are used in a cluster. The lack of this info will cause the monitoring-operator app to redeploy indefinitely, preventing migration.

Also ref: https://github.com/rancher/rancher/issues/32416
Also ref: https://github.com/bashofmann/rancher-monitoring-v1-to-v2#check-why-monitoring-v1-is-not-disabled

### DIFF
--- a/content/rancher/v2.5/en/monitoring-alerting/migrating/_index.md
+++ b/content/rancher/v2.5/en/monitoring-alerting/migrating/_index.md
@@ -46,7 +46,14 @@ Before you can install Monitoring V2, Monitoring V1 needs to be uninstalled comp
 * Disable the cluster monitoring installation under Cluster -> Tools -> Monitoring
 * Ensure that the cluster-monitoring app and the monitoring-operator app in the System project have been removed and are not recreated after a few minutes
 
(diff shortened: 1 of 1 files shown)


krumware:
despite the information here, the variables suggested to not seem to update the clusterSpec. When redeployed after being updated, the cluster spec still appears to show `"enableClusterAlerting": true,
"enableClusterMonitoring": true,`, even though there are no alerts, alert groups, notifiers, or other related monitoring resources

catherineluse:
@aiyengar2 In this PR, shouldn't the values be set to false to disable monitoring v1?

krumware:
@catherineluse You're absolutely right, I'll update the branch!

krumware:
@aiyengar2 I think you got pulled into this unintentionally, but do you have any input on whether it would be appropriate to remove those values from the template altogether, or is setting them to `false` acceptable?  Thanks for your help!

aiyengar2 (approved):
I'm not too familiar with Monitoring V1 (or with deploying it via RKE templates), but I think this looks fine to me as long as we have tested that this approach works. Setting it to false explicitly seems like the right approach to me.

PR #3291, opened 2021-05-25
Updating example of "Customizing your Ingress"

rtgibbons:
This is a bit deceiving in the current form has the path for cert-manager has changed. Other changes include:

  *  the changing the value to more identifiable suggestion
  * Adding ingress.tls.source to secret which has the template skip the setting of Issuer for the ingress, allowing ClusterIssuer to take effect.


superseb (changes requested):
@rtgibbons The path makes sense, it was changed upstream so we should also change it. The other part is unclear, the chart seems to set the annotations if it is not set to `secret`, and it also sounds like a workaround for what should be a config option in the chart to skip setting the annotations (or detect that cluster-issuer is set)

btat (approved):

PR #3294, opened 2021-05-26
Update snapshot to false for RKE cluster.yml example

dkeightley:
The snapshot: true refers to the legacy recurring snapshot syntax (pre RKE v0.2.0)
  * https://rancher.com/docs/rke/latest/en/etcd-snapshots/recurring-snapshots/


catherineluse:
I've made this change to the versioned docs https://github.com/rancher/docs/commit/f427dada8df1545950b2794682eb6dc12f525fce

PR #3302, opened 2021-05-26
Update helm install docs to reflect cluster size/replicas considerations

brandond:
Running 3 replicas on a 1 or 2 node cluster doesn't provide any benefit. Since we're not willing to change the default in the helm chart, we should instruct users on how to do this for themselves.

### DIFF
--- a/content/rancher/v2.5/en/installation/install-rancher-on-k8s/_index.md
+++ b/content/rancher/v2.5/en/installation/install-rancher-on-k8s/_index.md
@@ -172,18 +172,20 @@ The exact command to install Rancher differs depending on the certificate config
 {{% tab "Rancher-generated Certificates" %}}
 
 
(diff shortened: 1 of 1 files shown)


catherineluse:
Thanks for the PR Brandon. Can you please make the changes to the `content/rancher/v2.5` folder? The 2.x docs aren't maintained anymore.

catherineluse:
Also, do these changes apply only to v2.5.8+?

brandond:
Sure thing. Is there something that needs to change to get the "edit this page" link to point to the right branch? I just started working on the page that link pointed me at.

brandond:
And no, apparently the helm chart has always defaulted to 3 replicas but nobody could ever agree on whether that was safe to change.

catherineluse (approved):

Issue #3297, opened 2021-05-26
Add docs on nginx ingress port options


superseb:
**Summary:**
The options added in https://github.com/rancher/rke/issues/1876#issuecomment-716757170 should be added to https://rancher.com/docs/rke/latest/en/config-options/add-ons/ingress-controllers/

PR #3305, opened 2021-05-27
Describe ingress controller network options

superseb:
https://github.com/rancher/docs/issues/3297


aSauerwein:
i tried to configure ingress-nginx with metallb loadbalancer IP today in my lab and found out that "publish-service" extra also needs to be set for the ingress ressource. 
otherwise all the ingress objects will have the node ip as address, which causes issues when using external-dns

this is my config which solves all that 
```yaml
  ingress:
    default_backend: false
    extra_args:
      http-port: '80'
      https-port: '443'
      publish-service: $(POD_NAMESPACE)/ingress-nginx-controller
    http_port: 0
    https_port: 0
    network_mode: none
    provider: nginx
   ```

PR #3315, opened 2021-05-31
Include CSR example

ansilh:
To fix https://github.com/rancher/rke/issues/2550

### DIFF
--- a/content/rke/latest/en/cert-mgmt/_index.md
+++ b/content/rke/latest/en/cert-mgmt/_index.md
@@ -108,3 +108,16 @@ INFO[0001] [certificates] Generating etcd-zzzzz certificate and key
 INFO[0001] Successfully Deployed state file at [./cluster.rkestate]
 INFO[0001] Rebuilding Kubernetes cluster with rotated certificates
 ```
+
+### Optional settings 
+
+Use [extra_args]({{<baseurl>}}rke/latest/en/config-options/services/services-extras/#extra-args) to enable optional settings of each [services]({{<baseurl>}}/rke/latest/en/config-options/services/) that needs certificates and keys.
+
+For example, to enable [Certificate Management](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/), add `cluster-signing-cert-file` & `cluster-signing-key-file` to kube-controller in `cluster.yml` file.
+```
+services:
+  kube-controller: 
+    extra_args: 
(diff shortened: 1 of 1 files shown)


superseb:
These are two different things and should probably be on a different page, and if on the same page, it should be made more clear what Certificate Management means as this page describes how RKE handles it's certificates and that is unrelated to k8s certificate management.

ansilh:
Search engines will take the user to the certificate management doc page of RKE, and I thought it would be a good place to indicate it.
Can we add a sentence like the below to differentiate the cert management of both rke and k8s?
```
The Certificate Management mentioned in this example is unrelated to the certificate management of RKE.
Please refer to [Kubernetes documentation](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/) for more details on certificate management by Kubernetes.
```
Or is there any other good place to add this info?

superseb:
The user is unaware of what is what, currently it is specified as `Optional settings` but they are both very different settings. One is RKE's internal certificate management used for the k8s components, the other is a setting in k8s to manage certificates, which is just plain/vanilla Kubernetes as any other distro. We describe in our docs what RKE is capable of, not everything that k8s is capable of.

If there is confusion (which is logical if you end up on this page), just adding `Optional settings` with an example is not going to help the user. The RKE part needs to be clarified and the k8s part needs to be added in a way that its clear that it is a k8s thing and that you can enable it using the following setting.

ansilh:
So the conclusion would be; 
- RKE documentation has `extra_args` documented for enabling k8s specific features. 
- [Kubernetes documentation](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/#a-note-to-cluster-administrators) has the details about the flag needed to use a specific feature.
The use should be knowing both.

If that is the case, then we can close this PR.

superseb:
If people are looking this up and end up on this page, we can add a notice to say this is not k8s certificate management and link to another page where we show how to configure it. At least they are pointed in the right direction that way.

ansilh:
Sure. A new page or an existing one?
( https://rancher.com/docs/rke/latest/en/example-yamls/ ? )

superseb:
Yep the full example is fine, we don't need a separate page as we have k8s docs + the example yaml which will include the example

ansilh:
Updated the note and opened a PR https://github.com/rancher/docs/pull/3412 to add the details in the example cluster.yaml

Issue #3312, opened 2021-05-28
[Rancher2] 
Make the change from this PR in the versioned docs

https://github.com/rancher/docs/pull/3311


catherineluse:
**Request Summary:**

**Details:**

https://github.com/rancher/docs/pull/3311

PR #3316, opened 2021-05-31
Add clusterRole and binding to use restricted psp to versioned docs

btat:
Closes https://github.com/rancher/docs/issues/3312


catherineluse:
Thanks Billy!

PR #3380, opened 2021-07-01
Update steps to add cluster

kolonuk:
Documentation doesn't reflect new-user experience - took me ages to find out how to add a new cluster following the supposed "quick-start" guide!!

errors if we deleted it.

### DIFF
--- a/content/rancher/v2.5/en/quick-start-guide/deployment/quickstart-manual-setup/_index.md
+++ b/content/rancher/v2.5/en/quick-start-guide/deployment/quickstart-manual-setup/_index.md
@@ -46,7 +46,7 @@ To install Rancher on your host, connect to it and then use a shell to install.
 
 1.  Log in to your Linux host using your preferred shell, such as PuTTy or a remote Terminal connection.
 
-2.  From your shell, enter the following command:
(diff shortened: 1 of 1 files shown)


catherineluse:
What version of Rancher are you using?

btat:
@jtravee: I replaced your commit with updated steps since your updated was based on the original contributor's commit which wasn't fully accurate (terminology Cluster Dashboard vs Explorer, missing step that was the source of confusion, formatting, etc). Please review the updated version.

Additional context: When this was originally submitted I tried following the steps in our quickstart and wasn't able to reproduce the issue the contributor was encountering where the Cluster Explorer would be the landing page (which is why we tried asking the version above to get more info on repro steps, but got no response). I did more testing just now and it turns out there's a step that's not mentioned in the docs (or the original PR commit) where there's an option to set the default view during the initial log in. This option affects whether the Cluster Manager or Cluster Explorer is displayed. Also while testing this, when first clicking the "Cluster Manager" button you don't need to switch to the "Global" view whereas subsequent clicks on the "CM" button while in the CE view will require switching from "local" to "Global".

jtravee (approved):

PR #3384, opened 2021-07-06
Update _index.md

chrisludwig:
With the current link you end up (inapproproiate redirect in the background?) here: https://rancher.com/docs/rancher/v2.5/en/faq/removing-rancher/
Actual content one is looking for: https://rancher.com/docs/rancher/v2.5/en/cluster-admin/cleaning-cluster-nodes/


btat:
LGTM, thanks!

PR #3387, opened 2021-07-07
Update _index.md

nscarlettwolf:
Removing single quotes from EOF here doc delimiter.  With the quotes present, the ${proxy_host} variable is not substituted during the creation of /etc/apt/apt.conf.d/proxy.conf.  This leads to errors during "apt update" such as "Could not resolve '${proxy_host}'".

errors if we deleted it.


btat:
Woops. Sorry, I didn't realize separate PRs were made to fix the same issue so I applied this change to your other PR already and merged that.

PR #3388, opened 2021-07-07
remove single quotes from 'EOF' here doc delimiter

nscarlettwolf:
Removing single quotes from EOF here doc delimiter for command creating docker.service.d/http-proxy.conf.  With the quotes present, the variable ${proxy_host} is not substituted.

errors if we deleted it.


btat:
Thanks for the submission. There was another instance of the same issue which I've updated.

PR #3409, opened 2021-07-21
Fix networkpolicy for traefik ingress

rancher-max:
Fixing the network policy in the hardening guide for both traefik and traefik-svclb to policies that work for both k3s 1.20 and 1.21

https://github.com/k3s-io/k3s/issues/3583

### DIFF
--- a/content/k3s/latest/en/security/hardening_guide/_index.md
+++ b/content/k3s/latest/en/security/hardening_guide/_index.md
@@ -302,24 +302,50 @@ spec:
   - Ingress
 ```
 
-If you are using the default traefik ingress controller with k3s, it will also be blocked by default, so the following network policy must be added to allow traffic to both traefik pods and svclb pods in the kube-system namespace:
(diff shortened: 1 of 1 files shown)


content/k3s/latest/en/security/hardening_guide/_index.md
```diff
@@ -302,24 +302,50 @@ spec:
   - Ingress
 ```
 
-If you are using the default traefik ingress controller with k3s, it will also be blocked by default, so the following network policy must be added to allow traffic to both traefik pods and svclb pods in the kube-system namespace:
+If you are using the default traefik ingress controller with k3s, it will also be blocked by default, so the following network policies must be added to allow traffic to both traefik pods and svclb pods in the kube-system namespace. For version 1.20 and below there is a different label used than in 1.21 and above, so remove the one that is not associated with your kubernetes version.
```
briandowns, line 305:
```suggestion
If you are using the default traefik ingress controller with k3s, it will also be blocked by default, so the following network policies must be added to allow traffic to both traefik pods and svclb pods in the kube-system namespace. For version 1.20 and below there is a different label `traefik` used than in 1.21 and above, so remove the one that is not associated with your Kubernetes version.
```

briandowns (approved):

galal-hussein (approved):

brandond (approved):

PR #3436, opened 2021-08-11
Added resource useage best practices

Tejeev:
We're still seeing lots of OOM events from WAL compaction, and these defaults are not good for all environments. I have attempted to address this in the docs by adding a best practice of tuning resource limits and mentioning the issue in known issues.


Tejeev:
It's odd to me that there is a "known issue" to increase the Prometheus memory limits to those noted in the table above...

Tejeev:
After looking at https://github.com/rancher/charts/blob/main/charts/rancher-monitoring/values.yaml, I believe the `prometheus.prometheusSpec.resources` noted in that known issue refers to the `Prometheus` Resource in the table.

image

PR #3437, opened 2021-08-11
added docs for bootstrap password

paynejacob:

errors if we deleted it.


catherineluse:
Sorry, I should have clarified. The changes need to be made in the v2.6 directory, not the 2.x directory.

catherineluse (approved):

PR #3443, opened 2021-08-13
Clarified wording on authorized cluster endpoints

### LINKED ISSUE #3415: [Rancher2] Wording needs to be more explicit
Feedback from Support and a customer has indicated the excerpt below would benefit from being reworded to make things more explicit.

```
The authorized cluster endpoint only works on Rancher-launched Kubernetes
clusters. In other words, it only works in clusters where Rancher used RKE to
provision the cluster. It is not available for clusters in a hosted Kubernetes
provider, such as Amazon's EKS.
```

https://rancher.com/docs/rancher/v2.5/en/cluster-provisioning/rke-clusters/options/#authorized-cluster-endpoint

https://rancher.slack.com/archives/C6U4L2DS9/p1627479110193900

jtravee:
Closes #3415


catherineluse:
I agree with Billy. Basically we have different levels of management abilities for imported clusters (which are also sometimes called registered clusters), hosted clusters, and Rancher-provisioned clusters. Can we change

> cannot be enabled nor imported into Rancher from an RKE cluster;

to:

> cannot be enabled for RKE clusters that are imported into Rancher

And can you please also update this table https://rancher.com/docs/rancher/v2.5/en/cluster-provisioning/#cluster-management-capabilities-by-cluster-type to have another row for Authorized Cluster Endpoint, where there is only one checkmark, just for Rancher launched clusters?

catherineluse (approved):

btat (approved):

Issue #3448, opened 2021-08-18
[Rancher2] Document GKE cluster management enhancements


catherineluse:
- **Network tags for GKE clusters:** When provisioning node pools in GKE clusters, users can add [network tags](https://cloud.google.com/vpc/docs/add-remove-network-tags) to make firewall rules and routes between subnets.
- **Project Network Isolation for registered GKE clusters**: Admins and project owners can now leverage project network isolation to restrict network access between projects on imported/registered GKE clusters.

PR #3455, opened 2021-08-19
Add PNI and network tag to GKE options

btat:
Closes https://github.com/rancher/docs/issues/3448


catherineluse:
At the top of the GKE config reference can we also have a section for 'Changes in Rancher v2.6' with the two items?

btat:
> 
> 
> At the top of the GKE config reference can we also have a section for 'Changes in Rancher v2.6' with the two items?

Added

PR #3457, opened 2021-08-20
[fix] logging docu: splunk config and indentation of example code

papanito:
The splunk configuration was wrong as `splunkHec` starts with lower `s`.
Also fields are `hec_host` and `hec_port`.

errors if we deleted it.

### DIFF
--- a/content/rancher/v2.5/en/logging/custom-resource-config/outputs/_index.md
+++ b/content/rancher/v2.5/en/logging/custom-resource-config/outputs/_index.md
@@ -122,13 +122,13 @@ Let's say you wanted to send all logs in your cluster to an `elasticsearch` clus
 apiVersion: logging.banzaicloud.io/v1beta1
 kind: ClusterOutput
 metadata:
-    name: "example-es"
-    namespace: "cattle-logging-system"
+  name: "example-es"
+  namespace: "cattle-logging-system"
 spec:
-    elasticsearch:
-      host: elasticsearch.example.com
-      port: 9200
-      scheme: http
+  elasticsearch:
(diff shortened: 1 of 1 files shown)


ansilh:
The doc on the banzai side is [autogenerated vi acode](https://github.com/banzaicloud/logging-operator/blob/master/pkg/sdk/model/output/splunk_hec.go) and looks like the mistake started from there.

papanito:
Ahh ok so I close this PR and make changes there!?

ansilh:
Let's merge this PR and keep Rancher docs accurate. 
Sure, you can submit a PR in logging operator upstream too.

papanito:
Yeah I also fixed it upstream:
- https://github.com/banzaicloud/logging-operator-docs/pull/80
- https://github.com/banzaicloud/logging-operator/pull/805

ansilh:
Thanks @papanito 
Was there any reason not to change `splunkHec` (starting with small "s" instead of caps) in the upstream?

papanito:
Nope I guess I missed that. Will fix it...

papanito:
@catherineluse anything missing? 
Btw. I assume now that you have 2.6 doxu it should also go there?

catherineluse:
@papanito Yes, if it's valid for Rancher 2.6 it should go in the 2.6 directory.

papanito:
I will create a separate PR for 2.6.x but I guess would not hurt to also update 2.5.x docu ;-)

btat:
Merging and will use https://github.com/rancher/docs/issues/3520 for tracking of the 2.6 port

papanito:
Thanks @btat

Issue #3460, opened 2021-08-20
[Rancher2] New Steps for Rolling Back Rancher v2.5.0+


jiaqiluo:
Related Issue: https://github.com/rancher/rancher/issues/34247

**Request Summary:**

The existing steps on [this page for rolling back Rancher in 2.5.0+](https://rancher.com/docs/rancher/v2.5/en/installation/install-rancher-on-k8s/rollbacks/#rolling-back-to-rancher-v2-5-0) is not correct anymore. and we need to update it to the following:

step 1: find the deployment Rancher ( either in UI or command line) to scale it down to replica=0.
step 2: restore the backup, which is the same as the existing section [Create the Restore Custom Resource](https://rancher.com/docs/rancher/v2.5/en/installation/install-rancher-on-k8s/rollbacks/#create-the-restore-custom-resource)
step 3: run Helm rollback command to roll back rancher to its previous version. 
```
> helm rollback rancher -n cattle-system
```
Note: run `helm history rancher -n cattle-system` to determine which revision you should roll back to if the previous revision is not the target one; run `helm rollback -h` for help.

btat:
Closed via #3477

PR #3477, opened 2021-08-30
Update steps for rolling back v2.5+

btat:
Closes https://github.com/rancher/docs/issues/3460


content/rancher/v2.6/en/installation/install-rancher-on-k8s/rollbacks/_index.md
```diff
@@ -50,7 +60,7 @@ A restore is performed by creating a Restore custom resource.
 
 1. Click **Create**.
 
-**Result:** The rancher-operator scales down the rancher deployment during restore, and scales it back up once the restore completes. The resources are restored in this order:
+**Result:** The rancher-operator scales the rancher deployment back up once the restore completes. The resources are restored in this order:
```
jiaqiluo, line 63:
```suggestion
**Result:** The backup file is created and updated to the target storage location. The resources are restored in this order:
```

Explanation: Because we manually set the replica to 0 before creating the backup, so what the rancher-operator does is to scale down from 0 to 0 -> make the backup -> scale up from 0 to 0.

content/rancher/v2.6/en/backups/restoring-rancher/_index.md
```diff
@@ -8,7 +8,17 @@ A restore is performed by creating a Restore custom resource.
 > **Important**
 >
 > * Follow the instructions from this page for restoring rancher on the same cluster where it was backed up from. In order to migrate rancher to a new cluster, follow the steps to [migrate rancher.]({{<baseurl>}}/rancher/v2.6/en/backups/migrating-rancher)
```
jiaqiluo, line 10: All changes in this file can be dropped. The new steps are for the rollback case.
To be clear: 
restore == keep Rancher version the same, but restore its data from a backup file.
rollback == restore the backup made when Rancher runs an older version, and also rollback Rancher to the older version.

jiaqiluo (changes requested):

Issue #3343, opened 2021-06-17
Document configuration options in the Rancher UI for K3s and RKE2 clusters


catherineluse:

btat:
Closed via #3481

Issue #2900, opened 2020-12-10
[rke2] RKE2 -> Rancher 2.6 Integration


davidnuzik:
**Summary:**

In Rancher 2.6, additional support for RKE2 will be implemented. We currently plan to:
- Be able to provision RKE2 clusters in new Vue UI.
- Be able to support most or all the same CNI as RKE1 today
- Support Cluster templates
- Support config of nodes
- Support etcd backup and restore and rollback*
- Support addons
- Support CIS 1.6 in Rancher for RKE2
- Upgrades
- Airgap

In essence we should be very close to feature parity with RKE1. Everything I call out above may not need explicit documentation but I thought it would help to call these out. We may need to discuss further later to flush all this out.

**Details:**

catherineluse:
@davidnuzik can you change all the same options for registered RKE2 clusters that you can for clusters created within Rancher?

davidnuzik:
@catherineluse terribly sorry for the very late response. RKE2 won't have complete 100% feature parity, no. Happy to ask a dev on k3s/rke2 team to help identify differences if you would like. Just let me know.

catherineluse:
Confirmed flannel/weave aren't supported for RKE2.

btat:
Related issues:
- [ ] https://github.com/rancher/docs/issues/3258
- [x] https://github.com/rancher/docs/issues/3326
- [x] https://github.com/rancher/docs/issues/3343
- [ ] https://github.com/rancher/docs/issues/3369
- [x] https://github.com/rancher/docs/issues/3450
- [x] https://github.com/rancher/docs/issues/3463
- [x] https://github.com/rancher/docs/issues/3466

PR #3481, opened 2021-08-30
Add RKE2/K3s config options

btat:
#3343, #2900


btat:
Merging first for easier readability on staging site. Will address any feedback afterwards.

PR #3499, opened 2021-09-08
Documented default SUC deployment with Rancher k3s upgrades

### LINKED ISSUE #3190: [K3s] Document default SUC deployment with Rancher k3s upgrades
**Request Summary:**
When changing the k3s version for a cluster in the Rancher UI/API, the system upgrade controller is automatically deployed as a convenience and manages the plans to perform the change. The docs currently suggest the SUC is deployed manually, which can be confusing if a mixture of upgrade options are used.

For eg, an upgrade in the Rancher UI is performed, later an upgrade is performed by rolling in a new k3s binary, in this scenario the new k3s binary is unexpectedly downgraded to match the version stored in Rancher and the plan.

This may better suit the Rancher docs, but relates to k3s.

gz#16344

jtravee:
Closes #3190

### DIFF
--- a/content/k3s/latest/en/upgrades/automated/_index.md
+++ b/content/k3s/latest/en/upgrades/automated/_index.md
@@ -24,14 +24,11 @@ For more details on the design and architecture of the system-upgrade-controller
 - [system-upgrade-controller](https://github.com/rancher/system-upgrade-controller)
 - [k3s-upgrade](https://github.com/rancher/k3s-upgrade)
 
-To automate upgrades in this manner you must:
-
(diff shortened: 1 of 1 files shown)


dweomer (commented):
We should leave the steps as-is but we should also make an early call-out that if you have Rancher managing your cluster that the controller + plan installation/configuration steps are taken care of for you by Rancher.

cjellick:
@dweomer can you ciricle back to this? If youre feedback is addressed, can you officially approve?

dweomer (approved):

PR #3523, opened 2021-09-16
Update _index.md

wrkode:
corrected the --restart policy syntax


btat:
@wrkode are you encountering issues with the current syntax? Looking at the [docs](https://docs.docker.com/engine/reference/commandline/run/#restart-policies---restart), the current syntax is accepted as well.

PR #3538, opened 2021-09-22
Port missing in mirrors example

julien235:
Without port specify the private registry is unreachable.
I found solution here : https://github.com/k3s-io/k3s/issues/1713#issuecomment-742398187


brandond:
This isn't necessary. The example configuration allows you to refer to images on your private registry as `mycustomreg.com/image:tag`. If you refer to your images as `mycustomreg.com:5000/image:tag` then no specific endpoint configuration is necessary as the correct port is included in the image reference itself.

julien235:
@brandond : Yes indeed ! in my case I use http scheme.
Thanks for your explanation I close this PR.

PR #3539, opened 2021-09-22
Documented patch for Windows containers to be supported on Windows Server Standard Edition

### LINKED ISSUE #3536: Document patch for Windows containers to be supported on Windows Server Standard Edition
From https://github.com/rancher/windows/issues/72

Reference: https://rancher.com/docs/rancher/v2.6/en/cluster-provisioning/rke-clusters/windows-clusters/

jtravee:
Closes #3536

### DIFF
--- a/content/rancher/v2.6/en/cluster-provisioning/rke-clusters/windows-clusters/_index.md
+++ b/content/rancher/v2.6/en/cluster-provisioning/rke-clusters/windows-clusters/_index.md
@@ -37,6 +37,7 @@ Rancher v2.6 introduces provisioning for [RKE2](https://docs.rke2.io/) clusters 
 The RKE2 provisioning tech preview also includes installing RKE2 on Windows clusters. Windows features for RKE2 include:
 
 - Windows Containers with RKE2 powered by containerd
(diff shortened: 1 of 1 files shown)


content/rancher/v2.6/en/cluster-provisioning/rke-clusters/windows-clusters/_index.md
```diff
@@ -57,6 +58,8 @@ For the support lifecycle dates for Windows Server, see the [Microsoft Documenta
 
 Kubernetes v1.15+ is required.
 
+> **Note:** Kubernetes v1.21 will now work with Windows 20H2 Standard by installing the patch "2019-08 Servicing Stack Update for Windows Server."
```
btat, line 61:
```suggestion
If you are using Kubernetes v1.21 with Windows Server 20H2 Standard Core, the patch "2019-08 Servicing Stack Update for Windows Server" must be installed on the node.
```

The "will now work" part isn't completely accurate because it would have worked before if we had the patch installed, but we didn't know about that requirement.

btat (changes requested):

PR #3560, opened 2021-09-28
Update advanced TLS settings page

bashofmann:
This adds a detailed description on how to configure tls version and ciphers when Rancher is installed in a Kubernetes cluster. TLS is offloaded at the ingress controller, which means these settings need to be set there instead of the Rancher Pod/Container.

For running Rancher as a single Docker container, the default values have been updated to the ones currently supported.


btat:
@bashofmann are these changes applicable to 2.5 as well?

bashofmann:
@btat Yes, they are. Should I update them as well?

btat:
@bashofmann Yes, please update 2.5 as well. Thanks!

bashofmann:
@btat done

PR #3563, opened 2021-09-29
Harvester Tech Preview page added

jtravee:
https://github.com/rancher/docs/issues/3496

Reference new page: https://rancher.com/docs/rancher/v2.6/en/virtualization-admin/

PR #3565, opened 2021-09-30
Added note on Harvester node drivers

jtravee:
Harvester Tech Preview reference #3563.

### DIFF
--- a/content/rancher/v2.6/en/admin-settings/drivers/node-drivers/_index.md
+++ b/content/rancher/v2.6/en/admin-settings/drivers/node-drivers/_index.md
@@ -7,6 +7,8 @@ Node drivers are used to provision hosts, which Rancher uses to launch and manag
 
 If there are specific node drivers that you don't want to show to your users, you would need to de-activate these node drivers.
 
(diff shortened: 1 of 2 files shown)


content/rancher/v2.6/en/admin-settings/drivers/node-drivers/_index.md
```diff
@@ -7,6 +7,8 @@ Node drivers are used to provision hosts, which Rancher uses to launch and manag
 
 If there are specific node drivers that you don't want to show to your users, you would need to de-activate these node drivers.
 
+Harvester v0.3.0 is in tech preview for Rancher v2.6.1. For more information on Harvester node drivers, click [here](https://rancher.com/docs/rancher/v2.6/en/virtualization-admin/#harvester-node-driver).
```
btat, line 10: Could we add Harvester to the list on content/rancher/v2.6/en/admin-settings/drivers/_index.md instead of here.

btat (changes requested):

content/rancher/v2.6/en/admin-settings/drivers/node-drivers/_index.md
jtravee: Moved as suggested. Since the default is off rather than on, I didn't add to the list (which are all on by default), but added a new comment under list.

PR #3566, opened 2021-09-30
Merge master to staging

btat:


btat:
Forgot to change base to staging. Reverting in https://github.com/rancher/docs/pull/3567

PR #3580, opened 2021-10-01
Clarify that K3S_TOKEN must not be in K10 format

dcermak:


content/k3s/latest/en/installation/install-options/server-config/_index.md
```diff
@@ -59,7 +59,7 @@ In this section, you'll learn how to configure the K3s server.
 
 | Flag | Environment Variable | Description |
 |------|----------------------|-------------|
-|  `--token value, -t` value  | `K3S_TOKEN`      | Shared secret used to join a server or agent to a cluster |
+|  `--token value, -t` value  | `K3S_TOKEN`      | Shared secret used to join a server or agent to a cluster. This token must **not** be in the `K10` format (i.e. it starts with the string `K10`), use a random string instead. |
```
dereknola, line 62:
```suggestion
|  `--token value, -t` value  | `K3S_TOKEN`      | Shared secret used to join a server or agent to a cluster. Using the `K10` format is not recommended as additional configuration is required. |
```

dereknola: I suggest that we not be explicit, because using K10 formatting is possible, it just requires extra steps. I will open up an issue to document this process that but is essential boils down to:
- K10 format is used we treat the token as `K10<ca-hash>::<username>:<password>`. If the K10 prefix is missing we treat the token as just `<password>`.
- To use the K10 format you would need to:
1- start server with --token=randomstring
2- get the full K10 token off the server
3- use full K10 token to join agents and other servers

So its not that you can't use K10 it is just not the go-to basic setup.

dereknola (changes requested):

Issue #3079, opened 2021-03-02
[Rancher2] 


paulchoi:
**Request Summary:**

Clarify that the `s3-creds` Secret resource needs base64-encoded access key and secret key. While it's normally expected that the data in a Secret resource should be base64-encoded, it's easy to miss that detail. Please make it more obvious.

In https://github.com/rancher/docs/blob/master/content/rancher/v2.x/en/backups/v2.5/migrating-rancher/_index.md

**Details:**

Change as such:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: s3-creds
type: Opaque
data:
  accessKey: <Enter your base64-encoded access key>
  secretKey: <Enter your base64-encoded secret key>
```

PR #3589, opened 2021-10-05
Update _index.md

ealasgarov:

### DIFF
--- a/content/rancher/v2.5/en/backups/configuration/backup-config/_index.md
+++ b/content/rancher/v2.5/en/backups/configuration/backup-config/_index.md
@@ -129,6 +129,10 @@ s3:
 ```
 ### Example credentialSecret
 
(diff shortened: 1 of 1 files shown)


btat:
Hi @ealasgarov, are your secret and access keys first converted to base64 representation (as stated in your link) when getting the malformed error? There should be no difference whether the secret is created from a config file or directly with kubectl. If there is, then this would not be a documentation issue and the appropriate area to report this would be https://github.com/rancher/rancher/issues.

ealasgarov:
@btat Yes, you're right, probably my comment should be amended to: "convert the keys/secrets to base64" - this would suffice. This was not mentioned in the documentation and took me time to find out.

btat:
Hi @ealasgarov. thank you for the update (and PR). We've made it more clear now that those values should be base64-encoded so I will go ahead and close this one.

Ref: 
https://github.com/rancher/docs/pull/3594
https://github.com/rancher/docs/issues/3079

PR #3609, opened 2021-10-16
Update _index.md

yob8:
Updated command to create AKS cluster. Removed version restriction and added missing "\"

### DIFF
--- a/content/rancher/v2.5/en/installation/install-rancher-on-k8s/aks/_index.md
+++ b/content/rancher/v2.5/en/installation/install-rancher-on-k8s/aks/_index.md
@@ -48,9 +48,8 @@ To create an AKS cluster, run the following command. Use a VM size that applies 
 
 ```
 az aks create \
-  --resource-group rancher-rg
+  --resource-group rancher-rg \
   --name rancher-server \
-  --kubernetes-version 1.18.14 \
   --node-count 3 \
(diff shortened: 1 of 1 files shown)


content/rancher/v2.5/en/installation/install-rancher-on-k8s/aks/_index.md
```diff
@@ -48,9 +48,8 @@ To create an AKS cluster, run the following command. Use a VM size that applies
 
 ```
 az aks create \
-  --resource-group rancher-rg
+  --resource-group rancher-rg \
   --name rancher-server \
-  --kubernetes-version 1.18.14 \
```
btat, line 53: I think the version should be left in (but updated to 1.20.5) as each Rancher release has a highest version of AKS that's been validated/certified on based on our [support matrix](https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-5-9/). When the version isn't specified, the "cluster defaults to the N-1 minor version and latest patch" per the [AKS docs](https://docs.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#azure-portal-and-cli-versions), which could result in a version beyond what's been validated.

btat (changes requested):

Issue #3635, opened 2021-10-29
The 2.6 monitoring enablement guide refers to an option that does not exist


Tejeev:
Ref: https://rancher.com/docs/rancher/v2.6/en/monitoring-alerting/guides/enable-monitoring/
"Optional: Click Chart Options and configure alerting, Prometheus and Grafana. For help, refer to the configuration reference."

`Chart Options` is not an option.  Is this perhaps be referring to `Customize Helm options before install`?
![image](https://user-images.githubusercontent.com/8906089/139481965-793146b4-5e70-441d-9e18-621cd16788f6.png)

Tejeev:
After clicking it I see it leads to a very minor set of options.  From the description of the step, it looks like it's actually just talking about configuring the options exposed in `Install: Step 2` `Values` instead

Values:
![image](https://user-images.githubusercontent.com/8906089/139483379-facd599b-1946-46c3-a6ec-2e61f00f0282.png)
vs Helm Options:
![image](https://user-images.githubusercontent.com/8906089/139483430-bd346ba3-f24b-4809-8070-84d5b8390ce3.png)

btat:
Closed via https://github.com/rancher/docs/pull/3636.

PR #3636, opened 2021-10-29
I think these instructions were incorrect for 2.6

Tejeev:
Since the instructions appeared to reference steps in the path that doesn't exist when installing from Apps & Marketplace, I updated with a path that is shorter and correct for 2.6.x

Please note that I did not correct anything but the basic install instructions.  The other instructions may still need QA


Tejeev:
This follows on my filing of https://github.com/rancher/docs/issues/3635

PR #3333, opened 2021-06-13
Use latest instead of tag in Automated Upgrades

elmariofredo:

PR #3639, opened 2021-10-30
Update _index.md

itoffshore:
`v0.6.2` fails with a 'crashback loop' error - version `0.8.0` installs without issue

the suse.com page (where I found the hint) should probably also be changed:

https://www.suse.com/c/rancher_blog/upgrade-a-k3s-kubernetes-cluster-with-system-upgrade-controller/


btat:
Thank you for the PR. Looks https://github.com/rancher/docs/pull/3333 already handles this.

PR #3655, opened 2021-11-04
Add dual-stack warning

manuelbuil:

### DIFF
--- a/content/k3s/latest/en/installation/network-options/_index.md
+++ b/content/k3s/latest/en/installation/network-options/_index.md
@@ -72,6 +72,8 @@ You should see that IP forwarding is set to true.
 
 ### Dual-stack installation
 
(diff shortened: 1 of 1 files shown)


content/k3s/latest/en/installation/network-options/_index.md
```diff
@@ -72,6 +72,8 @@ You should see that IP forwarding is set to true.
 
 ### Dual-stack installation
 
+A dual-stack cluster must be configured when the cluster is first created. It cannot be enabled on an existing cluster.
```
brandond, line 75:
```suggestion
Dual-stack networking must be configured when the cluster is first created. It cannot be enabled on an existing cluster.
```

brandond (changes requested):

brandond (approved):

PR #3657, opened 2021-11-05
Added to note for v2.5 and v2.6 rollback docs clarifying that UI will be unavailable during the process 

### LINKED ISSUE #3642: [Rancher2] Add note to v2.5 and v2.6 rollbacks docs clarifying that UI will be unavailable during the process
**Request Summary:**
Add note to both the v2.5 and v2.6 rollbacks docs clarifying that the UI will be unavailable until the rollback process is completed. This is because scaling the `rancher` deployment pod down to `0` will result in both Rancher and its UI to not be in a running state.

https://rancher.com/docs/rancher/v2.5/en/installation/install-rancher-on-k8s/rollbacks/
https://rancher.com/docs/rancher/v2.6/en/installation/install-rancher-on-k8s/rollbacks/

**Details:**
cc @flrichar

jtravee:
Closes #3642

### DIFF
--- a/content/rancher/v2.5/en/installation/install-rancher-on-k8s/rollbacks/_index.md
+++ b/content/rancher/v2.5/en/installation/install-rancher-on-k8s/rollbacks/_index.md
@@ -26,7 +26,7 @@ A restore is performed by creating a Restore custom resource.
 > **Important**
 >
 > * Follow the instructions from this page for restoring rancher on the same cluster where it was backed up from. In order to migrate rancher to a new cluster, follow the steps to [migrate rancher.]({{<baseurl>}}/rancher/v2.5/en/backups/migrating-rancher)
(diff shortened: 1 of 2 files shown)


nickgerace (approved):

btat (approved):

wpwoodjr:
You might get an error like this when doing an `apply`:
```
$ kubectl apply -f restore.yaml 
error: from restore-: cannot use generate name with apply
```
Use `create` instead:
```
$ kubectl create -f restore.yaml 
restore.resources.cattle.io/restore-mppdg created
```

btat:
> 
> 
> You might get an error like this when doing an `apply`:
> 
> ```
> $ kubectl apply -f restore.yaml 
> error: from restore-: cannot use generate name with apply
> ```
> 
> Use `create` instead:
> 
> ```
> $ kubectl create -f restore.yaml 
> restore.resources.cattle.io/restore-mppdg created
> ```
> 
> @jtravee @nickgerace @btat

Thanks @wpwoodjr, it's been changed to use `create` now.

Ref: https://github.com/rancher/docs/pull/3667

PR #3659, opened 2021-11-06
Command added

mr-buryakov:
Command to change host name added


jtravee:
Closing as 1.x is only being maintained for critical issues at this time. Additionally, the `ros config` commands are listed under the link for the cloud config page already.

PR #3689, opened 2021-11-17
Add networkpolicy for metrics-server in hardened k3s

rancher-max:
Addresses https://github.com/k3s-io/k3s/issues/4496

### DIFF
--- a/content/k3s/latest/en/security/hardening_guide/_index.md
+++ b/content/k3s/latest/en/security/hardening_guide/_index.md
@@ -302,9 +302,23 @@ spec:
   - Ingress
 ```
 
-If you are using the default traefik ingress controller with k3s, it will also be blocked by default, so the following network policies must be added to allow traffic to both traefik pods and svclb pods in the kube-system namespace. For version 1.20 and below there is a different label `traefik` used than in 1.21 and above, so remove the one that is not associated with your Kubernetes version.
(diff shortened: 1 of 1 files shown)


paynejacob (approved):

content/k3s/latest/en/security/hardening_guide/_index.md
```diff
@@ -302,11 +302,25 @@ spec:
   - Ingress
 ```
 
-If you are using the default traefik ingress controller with k3s, it will also be blocked by default, so the following network policies must be added to allow traffic to both traefik pods and svclb pods in the kube-system namespace. For version 1.20 and below there is a different label `traefik` used than in 1.21 and above, so remove the one that is not associated with your Kubernetes version.
+A few of the packaged components may also need NetworkPolicies applied to allow them to work properly. For example, metrics-server will only work intermittently and the Traefik ingress controller will be blocked by default if not supplying the below policies. For Traefik in k3s version 1.20 and below there is a different `traefik` label used than in 1.21 and above, so remove the one in the below yaml that is not associated with your Kubernetes version.
```
brandond, line 305:
```suggestion
The metrics-server and Traefik ingress controller will be blocked by default if network policies are not created to allow access. Traefik v1 as packaged in K3s version 1.20 and below uses different labels than Traefik v2; ensure that you only use the sample yaml below that is associated with the version of Traefik present on your cluster.
```

brandond (changes requested):

brandond (approved):

PR #3712, opened 2021-12-02
Add missing steps for recovering from cert expiration

LinAnt:
Add instructions for deleting the mutating webhook, If this is not done, the rancher-webhook pod will not be able to recreate the certificate.

rancher-webhook will recreate the mutating webhook on startup :+1: 

### DIFF
--- a/content/rancher/v2.6/en/troubleshooting/expired-webhook-certificates/_index.md
+++ b/content/rancher/v2.6/en/troubleshooting/expired-webhook-certificates/_index.md
@@ -10,6 +10,7 @@ Rancher will advise the community once there is a permanent solution in place fo
 ##### 1. Users with cluster access, run the following commands:
 ```
 kubectl delete secret -n cattle-system cattle-webhook-tls
+kubectl delete mutatingwebhookconfigurations.admissionregistration.k8s.io rancher.cattle.io
 kubectl delete pod -n cattle-system -l app=rancher-webhook
 ```
 
@@ -17,6 +18,8 @@ kubectl delete pod -n cattle-system -l app=rancher-webhook
 
 1. Delete the `cattle-webhook-tls` secret in the `cattle-system` namespace in the local cluster.
 
-1. Delete the `rancher-webhook` pod in the `cattle-system` namespace in the local cluster.
+2. Delete the `rancher.cattle.io` mutating webhook
+
+3. Delete the `rancher-webhook` pod in the `cattle-system` namespace in the local cluster.
 
 **Note:** The webhook certificate expiration issue is not specific to `cattle-webhook-tls` as listed in the examples. You will fill in your expired certificate secret accordingly.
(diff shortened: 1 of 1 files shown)


content/rancher/v2.6/en/troubleshooting/expired-webhook-certificates/_index.md
```diff
@@ -10,13 +10,16 @@ Rancher will advise the community once there is a permanent solution in place fo
 ##### 1. Users with cluster access, run the following commands:
 ```
 kubectl delete secret -n cattle-system cattle-webhook-tls
+kubectl delete mutatingwebhookconfigurations.admissionregistration.k8s.io rancher.cattle.io
```
cbron, line 13: Appears to follow our pre-delete-hook job: https://github.com/rancher/webhook/blob/efe9b4b5743347554cc6c332d1b615de4c321b66/charts/rancher-webhook/templates/pre-delete-hook-job.yaml#L25

Adding @jiaqiluo to review since he has more context.

thedadams: Yes, I think the `--ignore-not-found=true` should be added, at the very least, for idempotency.

LinAnt:
```suggestion
kubectl delete mutatingwebhookconfigurations.admissionregistration.k8s.io --ignore-not-found=true rancher.cattle.io
```

LinAnt: Is there any reason the mutating webhook would be missing though? :thinking:

LinAnt:
There is this Note section in the bottom:

```
Note: The webhook certificate expiration issue is not specific to cattle-webhook-tls as listed in the examples. You will fill in your expired certificate secret accordingly.```

Which other certs should one renew? Also, renewal of those certs would likely also fail if the webhook cert expires at the same time.

content/rancher/v2.6/en/troubleshooting/expired-webhook-certificates/_index.md
thedadams: Yes. If a user ran the command, got distracted with other tasks and came back wondering where they left off. Making it easy/possible to run the command a second time is advantageous.

LinAnt: Ah, makes sense :+1:

thedadams (approved):
I verified in my local setup that these steps do allow the certificate to be regenerated.

jiaqiluo (approved):

jiaqiluo:
@affansyed  there should be a Deployment named `rancher-webhook` in the namespace `cattle-system` which will recreate the webhook pod after we delete it.  Could you check if that Deployment exists and if it works fine in your cluster?

PR #3823, opened 2022-01-25
Note regarding mega nodes

Tejeev:

### DIFF
--- a/content/rke/latest/en/os/_index.md
+++ b/content/rke/latest/en/os/_index.md
@@ -289,6 +289,14 @@ Confirm that a Kubernetes supported version of Docker is installed on your machi
 
 This section describes the hardware requirements for the worker role, large Kubernetes clusters, and etcd clusters.
 
+---
+*NOTE regarding large nodes for Kubernetes*
+Kubernetes is engineered around the concept of horizontal scaling for redundancy so scaling vertically with large nodes can be very problematic 
+- If you must use nodes larger than 24 CPU, utilize virtualization tooling such as Harvester to subdivide the nodes.
+- There are Kubernetes, kernel, and network limitations that prohibit too many pods per node so we advise keeping to roughly 24 cores per node and the recommended 100 pods per node as a maximum unless you are deploying workloads that specifically require enormous amounts of resources (such as multi threaded heavy compute jobs) or have good reason to increase the pod limit (with an upper limit of 250).
+- Even when deploying workloads like the above, it is recommended to use a virtualization layer to facilitate less downtime/shorter reboots during upgrades and failures
+---
+  
 ### Worker Role
 
 The hardware requirements for nodes with the `worker` role mostly depend on your workloads. The minimum to run the Kubernetes node components is 1 CPU (core) and 1GB of memory. 


btat (approved):

content/rke/latest/en/os/_index.md
```diff
@@ -297,7 +297,17 @@ Regarding CPU and memory, it is recommended that the different planes of Kuberne
 
 ### Large Kubernetes Clusters
 
-For hardware recommendations for large Kubernetes clusters, refer to the official Kubernetes documentation on [building large clusters](https://kubernetes.io/docs/setup/best-practices/cluster-large/).
+Kubernetes is engineered around the concept of horizontal scaling for redundancy, so scaling vertically with large nodes can be problematic if the proper minimum/maximum requirements aren’t followed. The following are tips and recommendations for large Kubernetes nodes:
+
+- If you must use nodes larger than 24 CPU, use virtualization tooling, such as what [Harvester](https://docs.harvesterhci.io/v1.0/rancher/virtualization-management/) provides, to subdivide the nodes.
+
+- Kubernetes, kernel, and network limitations prevent having too many pods per node. You should maintain a minimum of roughly 24 cores per node and a maximum of the recommended 100 pods per node.
```
dnoland1, line 304: Believe the recommended cap is 110, not 100 pods per node.

content/rke/latest/en/os/_index.md
```diff
@@ -297,7 +297,17 @@ Regarding CPU and memory, it is recommended that the different planes of Kuberne
...
+
+- If you are deploying workloads that specifically require enormous amounts of resources (such as multi-threaded heavy compute jobs), you may increase the pod limit up to 250.
```
dnoland1, line 306: If a workload, sometimes synonymous with a pod, is resource intensive, you would want less of them, not more. Maybe a better way to state this is:

If you are deploying an application or system that specifically requires a large number of pods, you may increase the pod limit up to 250.

Also, curious where the 250 suggestion comes from. Could also note that going above 254 is not supported by default pod CIDR settings unless pods are using host networking.

content/rke/latest/en/os/_index.md
```diff
@@ -297,7 +297,17 @@ Regarding CPU and memory, it is recommended that the different planes of Kuberne
 
 ### Large Kubernetes Clusters
 
-For hardware recommendations for large Kubernetes clusters, refer to the official Kubernetes documentation on [building large clusters](https://kubernetes.io/docs/setup/best-practices/cluster-large/).
+Kubernetes is engineered around the concept of horizontal scaling for redundancy, so scaling vertically with large nodes can be problematic if the proper minimum/maximum requirements aren’t followed. The following are tips and recommendations for large Kubernetes nodes:
+
+- If you must use nodes larger than 24 CPU, use virtualization tooling, such as what [Harvester](https://docs.harvesterhci.io/v1.0/rancher/virtualization-management/) provides, to subdivide the nodes.
```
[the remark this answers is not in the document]
jtravee: Updated per feedback.

content/rke/latest/en/os/_index.md
jtravee: Updated per feedback.

content/rke/latest/en/os/_index.md
jtravee: Removed the specificity of 250, and added your caveat, ty!

content/rke/latest/en/os/_index.md
```diff
@@ -297,7 +297,17 @@ Regarding CPU and memory, it is recommended that the different planes of Kuberne
 
 ### Large Kubernetes Clusters
 
-For hardware recommendations for large Kubernetes clusters, refer to the official Kubernetes documentation on [building large clusters](https://kubernetes.io/docs/setup/best-practices/cluster-large/).
+Kubernetes is engineered around the concept of horizontal scaling for redundancy, so scaling vertically with large servers can be problematic if the proper minimum/maximum requirements aren’t followed. The following are tips and recommendations for large Kubernetes clusters:
+
+- If you must use servers larger than 24 CPU, use virtualization tooling, such as what [Harvester](https://docs.harvesterhci.io/v1.0/rancher/virtualization-management/) provides, to subdivide the servers.
+
+- Kubernetes, kernel, and network limitations prevent having too many pods per server. You should maintain a minimum of roughly 24 cores per server and a maximum of the recommended 110 pods per server.
```
jtravee, line 304:
```suggestion
- Kubernetes, kernel, and network limitations prevent having too many pods per server. You should maintain a maximum of roughly 24 cores per server and a maximum of the recommended 110 pods per server.
```

jtravee: @Tejeev checking that `minimum` should be `maximum` here.

jtravee (changes requested):

content/rke/latest/en/os/_index.md
Tejeev: yes, though these are somewhat softer than I think I make it sound.  Folks that know what they are doing may up the pod count to the real limit of 251 (IIRC.  we should double check that) and there are use cases that might use more cores, per pod, though at the moment they are rare.  Not sure how to convey this a bit better or if we leave it as is to prevent over committing by folks that do not have real need of increased limits

jtravee (approved):

content/rke/latest/en/os/_index.md
Tejeev: Perhaps:
"It is recommended to maintain a maximum of roughly 24 cores and 110 pods per server"

content/rke/latest/en/os/_index.md
```diff
@@ -297,7 +297,17 @@ Regarding CPU and memory, it is recommended that the different planes of Kuberne
...
 
-For hardware recommendations for large Kubernetes clusters, refer to the official Kubernetes documentation on [building large clusters](https://kubernetes.io/docs/setup/best-practices/cluster-large/).
+Kubernetes is engineered around the concept of horizontal scaling for redundancy, so scaling vertically with large servers can be problematic if the proper minimum/maximum requirements aren’t followed. The following are tips and recommendations for large Kubernetes clusters:
+
+- If you must use servers larger than 24 CPU, use virtualization tooling, such as what [Harvester](https://docs.harvesterhci.io/v1.0/rancher/virtualization-management/) provides, to subdivide the servers.
+
+- Kubernetes, kernel, and network limitations prevent having too many pods per server. You should maintain a maximum of roughly 24 cores per server and a maximum of the recommended 110 pods per server.
+
+- If you are deploying an application or system that specifically requires a large number of pods, you may increase the pod limit. Please note, however, that going above 254 pods per server is not supported by default pod CIDR settings unless the pods are using host networking.
```
cbron, line 306: This goes against upstream docs best practice: _"No more than 110 pods per node"_: https://kubernetes.io/docs/setup/best-practices/cluster-large/

Also saying _"you may increase the pod limi"_ implies that this is supported, and we have not tested this in our scaling tests, we do not want to give this impression.

content/rke/latest/en/os/_index.md
```diff
@@ -297,7 +297,17 @@ Regarding CPU and memory, it is recommended that the different planes of Kuberne
 
 ### Large Kubernetes Clusters
 
-For hardware recommendations for large Kubernetes clusters, refer to the official Kubernetes documentation on [building large clusters](https://kubernetes.io/docs/setup/best-practices/cluster-large/).
+Kubernetes is engineered around the concept of horizontal scaling for redundancy, so scaling vertically with large servers can be problematic if the proper minimum/maximum requirements aren’t followed. The following are tips and recommendations for large Kubernetes clusters:
+
+- If you must use servers larger than 24 CPU, use virtualization tooling, such as what [Harvester](https://docs.harvesterhci.io/v1.0/rancher/virtualization-management/) provides, to subdivide the servers.
```
cbron, line 302: > If you must use servers larger than 24 CPU

Where did that 24 CPU number come from ?

content/rke/latest/en/os/_index.md
```diff
@@ -297,7 +297,17 @@ Regarding CPU and memory, it is recommended that the different planes of Kuberne
...
+
+- Kubernetes, kernel, and network limitations prevent having too many pods per server. You should maintain a maximum of roughly 24 cores per server and a maximum of the recommended 110 pods per server.
```
cbron, line 304: > roughly 24 cores per server 

Where did this data come from ?

content/rke/latest/en/os/_index.md
```diff
@@ -297,7 +297,17 @@ Regarding CPU and memory, it is recommended that the different planes of Kuberne
```
cbron, line 297: Is there a reason we put this in the rke docs ? Wouldn't this apply to any large cluster, including rke2. 
We need to make a scaling section, but for now most of that data lives over here: https://rancher.com/docs/rancher/v2.5/en/installation/requirements/

cbron (changes requested):
Given questions the scaling team has gotten in the past about our docs, we need to understand where any information comes from and how we got those numbers. That way we can understand and relay that information as necessary, as well as re-test in the future. More specifically if this is the average case, or a single case, what distro it references, if it's an upstream recommendation etc...

PR #3827, opened 2022-01-27
Switch order of 2 steps in "registring a cluster"

SafPlusPlus:
In my Rancher 2.6.3 install, when importing an existing RKE2 cluster, I needed to pick the type before entering the name of the new cluster. These 2 steps were documented in a different order.


btat (approved):
LGTM. Thanks!

PR #3834, opened 2022-02-01
The key under "mirrors" must match the endpoint

DracoBlue:
If you use "docker.io" you will still get:

```
http: server gave HTTP response to HTTPS client
```

like @danlister pointed out at https://github.com/k3s-io/k3s/issues/1713#issuecomment-621335992.

### DIFF
--- a/content/k3s/latest/en/installation/private-registry/_index.md
+++ b/content/k3s/latest/en/installation/private-registry/_index.md
@@ -60,7 +60,7 @@ Below are examples showing how you may configure `/etc/rancher/k3s/registries.ya
 
 ```
 mirrors:
-  docker.io:
+  "mycustomreg.com:5000":
     endpoint:
       - "https://mycustomreg.com:5000"
 configs:
@@ -79,7 +79,7 @@ configs:
 
 ```
 mirrors:
-  docker.io:
+  "mycustomreg.com:5000":
     endpoint:
       - "https://mycustomreg.com:5000"
 configs:
@@ -102,7 +102,7 @@ Below are examples showing how you may configure `/etc/rancher/k3s/registries.ya
 
 ```
 mirrors:
-  docker.io:
+  "mycustomreg.com:5000":
     endpoint:
       - "http://mycustomreg.com:5000"
 configs:
@@ -117,7 +117,7 @@ configs:
 
 ```
 mirrors:
-  docker.io:
+  "mycustomreg.com:5000":
     endpoint:
       - "http://mycustomreg.com:5000"
 ```


brandond:
The docs show how to use a local registry at `https://mycustomreg.com:5000` as a mirror for `docker.io`. Your change would remove this; the local registry would no longer be configured as a mirror for docker.io.

The error you're getting suggests that your registry is not using HTTPS. If you've configured your registry for plaintext HTTP then you should change the endpoint scheme to http:

```yaml
mirrors:
  docker.io:
    endpoint:
      - "http://mycustomreg.com:5000"
configs:
  "mycustomreg:5000":
    auth:
      username: xxxxxx # this is the registry username
      password: xxxxxx # this is the registry password
    tls:
      cert_file: # path to the cert file used in the registry
      key_file:  # path to the key file used in the registry
      ca_file:   # path to the ca file used in the registry
```

brandond (changes requested):
see comment

DracoBlue:
> The error you're getting suggests that your registry is not using HTTPS. If you've configured your registry for plaintext HTTP then you should change the endpoint scheme to http:

this is exactly what I did. But I had to replace docker.io with mycustomreg.com:5000 to make it work with http. Please see https://github.com/k3s-io/k3s/issues/1713#issuecomment-621335992 comment, because he solved the issue the same way.

I find it also missleading that you have to do this to make it available, but for the hostname mycustomreg.com:5000 as a key will be used to figure the `endpoint` to figure that it is http. Thus I expect that docker.io is wrong if you want to override the protocol for the hostname mycustomreg.com:5000.

IMHO it would be better if the mirrors content would not be reused to make the protocol changable, but instead have something like this implemented:

```
# this does not work, it's just an idea!
configs:
  "mycustomreg:5000":
    auth:
      scheme: "http"
```

DracoBlue:
But ... I get your point. If the documentation explains how to use a http or https private mirror, for docker.io - it needs to stay docker.io. But if you want to use mycustomreg:5000 with http as source for your images, none of the examples described in the "private registry".md file will work - thus my suggestion to change the page.

brandond:
If you want to pull `mycustomreg.com:5000/test-image` over http, then you need to use:
```
mirrors:
  "mycustomreg.com:5000":
    endpoint:
      - "http://mycustomreg.com:5000"
```
but that's not what this documentation is showing you how to configure. It's showing you how to configure contaienrd to pull from `mycustomreg.com:5000` as a mirror for `docker.io`.

DracoBlue:
So can we add how to use a "private registry" to store and pull images?

Because right now it explains how to use a local registry as mirror for docker.io, but only at the end that you can push to a private registry. I would like to get the information how to use the private registry as source for images (if running on http) on this page. This page is the first hit if you try to use private registries - but focuses on mirrors for docker.io :).

brandond:
> So can we add how to use a "private registry" to store and pull images?

You don't **need** to add anything to use a private registry. If you specify a Pod spec with `mycustomreg.com:5000/myimage` as the image, it will just work without any additional configuration. 

Both docker and containerd assume anonymous https with a valid cert, so manual configuration is only necessary if your registry:
* uses http instead of http instead of https
* does not have a valid cert
* requires authentication

If any of those are true, you would need to explicitly configure the endpoint, certs, auth, etc.

DracoBlue:
Yep. Sorry for the late response!

> So can we add how to use a "private registry" to store and pull images?

let me rephrase:

> So can we add how to use a "private non-tls registry" to store and pull images?

Maybe a section or sentence will do. I will close this PR to avoid confusion.

PR #3835, opened 2022-02-01
Update default admin password for Vagrant quickstart (v2.6 only)

thinkmassive:
Updates the v2.6 docs for local quickstart, to specify the new default admin password from [quickstart PR #202](https://github.com/rancher/quickstart/pull/202)

A quick look at rancher v2.5 makes me think the password length requirements only changed for v2.6+ If the 2.5 docs should also be updated, please let me know, or feel free to add a commit to this branch.


btat:
Thanks for the PR @thinkmassive. As you noted, the password length only applies to v2.6 so no changes are needed on the v2.5 docs.

PR #3842, opened 2022-02-02
Fix etcd Snapshot cron Expression

ShadowOfLies:
Unfamiliarity with cron expressions could result in bad configuration. In this case, the expression intended to represent every 5 hours, but instead represented every minute past every 5 hours (i.e. 60 snapshots every 5 hours)


brandond (changes requested):
This help text is generated by code and should also be updated there:
https://github.com/k3s-io/k3s/blob/df4147cd571f6135a965bded98f424450699d09b/pkg/cli/cmds/server.go#L282-L287

We'll need another issue to track that, though.

brandond (approved):
whoops, didn't look at the repo.

brandond (approved):

PR #3851, opened 2022-02-04
Update troubleshooting for Cannot Connect to Docker Daemon

hawkesn:
Adding a small update to the section: `Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?`
Both of these options work for me. Might help those who also run into this issue.


btat:
Hi @hawkesn , this repository is for [Rancher](https://github.com/rancher/rancher) documentation. Based on the contents of your PR I believe you're looking to contribute to the Rancher Desktop documentation which you can find at https://github.com/rancher-sandbox/docs.rancherdesktop.io. Please resubmit your PR there instead, thanks!

PR #3860, opened 2022-02-08
Mention Terraform more prominently in docs

catherineluse:
In the Rancher Users Slack, I've seen several people asking about how to programmatically deploy EKS clusters. The best way to do it is through Terraform, so I have mentioned that in more places that it is relevant and linked to the Terraform docs about creating clusters. This affects RKE1, EKS, AKS and GKE clusters.


btat (approved):
LGTM

PR #3890, opened 2022-02-21
Update AAD auth doc

ryanelliottsmith:
Include note on the fact that the user that is used to test authentication is bound to the 'admin' account.


content/rancher/v2.5/en/admin-settings/authentication/azure-ad/_index.md
```diff
@@ -206,3 +206,8 @@ Enter the values that you copied to your [text file](#tip).
 1. Click **Authenticate with Azure**.
 
 **Result:** Azure Active Directory authentication is configured.
+
+> **Note:**
+>
+> The AD user pertaining to the credentials used to test will be mapped to the local principal account and assigned administrator privileges in Rancher. You should therefore make a conscious decision on which account you use to perform this step.
+
```
cbron, line 213: Can you add this to all versions of the doc, 2.6, 2.4 etc.. 

Also, should this be done at a generic level for all oauth/ldap providers.

PR #3895, opened 2022-02-22
Use the current hostname for baseUrl

catherineluse:
Nelson said it would simplify the process of docs migration if we could have baseurl be whatever address the site is running on, instead of having to pass in the site name manually.

I have changed docker files for both dev and prod. The docker files no longer pass in any variables for the site name. Instead baseurl is now taken from the Hugo variable called `.Permalink`.

I have only tested this locally, as the links are currently pointing to local pages when the docs run locally. Next I would like these changes to be merged into staging so that we can confirm that links containing `baseUrl` are going to `https://staging.rancher.com` as expected. If that works as expected, this should be fine to merge to master.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -10,4 +10,4 @@ RUN mkdir -p /output /theme/rancher-website-theme && tar -xzf /run/master.tar.gz
 # Expose default hugo port
 EXPOSE 9001
 
-ENTRYPOINT ["hugo", "serve", "--bind=0.0.0.0", "--buildDrafts", "--buildFuture", "--baseURL=" ]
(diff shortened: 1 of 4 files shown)


divya-mohan0209 (commented):
This looks fine to me and I tested it with 4-5 links in different sections on my local machine. Should be good to be pushed to staging.
/approve

divya-mohan0209 (commented):
Will merge changes after call on Wednesday.

paraglade:
should this PR also include this Dockerfile: https://github.com/rancher/docs/blob/master/Dockerfile.staging#L19

catherineluse:
yes, good catch. I will update it.

PR #3899, opened 2022-02-22
Make docs version show up in search results

catherineluse:
The intention of this PR is to make the docs version show up in search results when you click the magnifying glass in the top nav of the docs.

This is a stopgap measure that is meant to improve the user experience in the short term. The long term solution is to migrate to docusaurus in which algolia search will be limited to only one docs version at a time.

Currently it shows identical entries in the list when they are really entries for two different docs versions.

I have been able to get something to show up below each list item, as shown in this screen shot:
Screen Shot 2022-02-22 at 2 51 17 PM

This PR is in progress because I haven't figured out how to get just the docs version in there. Currently I'm adding the URL that the search result list item is pointing to, which contains the version in it. I can use Javascript to extract the version from the URL itself, but I can't use javascript in this template which is used to render the list item: https://github.com/rancher/docs/blob/master/assets/js/app.js#L42 It seems those variables that are already in the template (such as `_highlightresult`, `_snippetresult` and `permalink`) have to be passed in to algolia in a special way. I find this Hugo layout file in the rancherlabs/website repo to be fascinating because it looks like it is used to put variables in the list results. https://github.com/rancherlabs/website-theme/blob/master/layouts/_default/list.algolia.json I found docs about how those variables work but I have yet to make sense of it. https://www.algolia.com/doc/api-reference/widgets/infinite-hits/js/

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -10,4 +10,4 @@ RUN mkdir -p /output /theme/rancher-website-theme && tar -xzf /run/master.tar.gz
 # Expose default hugo port
 EXPOSE 9001
 
-ENTRYPOINT ["hugo", "serve", "--bind=0.0.0.0", "--buildDrafts", "--buildFuture", "--baseURL=" ]
+ENTRYPOINT ["hugo", "serve", "--bind=0.0.0.0", "--buildDrafts", "--buildFuture" ]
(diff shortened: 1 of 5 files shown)


catherineluse:
@divya-mohan0209 I believe that to move forward on this, we will have to make sense of this doc. Maybe you can review it before we meet? https://github.com/rancher/docs/pull/3899

PR #3944, opened 2022-03-11
Fix no_proxy to noProxy for rancher chart

axeal:
value is noProxy (per https://github.com/rancher/rancher/blob/v2.6.3/chart/values.yaml#L89 and https://rancher.com/docs/rancher/v2.6/en/installation/install-rancher-on-k8s/chart-options/), this is correct in v2.5 doc at https://rancher.com/docs/rancher/v2.5/en/installation/other-installation-methods/behind-proxy/install-rancher/


jtravee:
Hi @axeal, could you also update this in the [Install Cert Manager](https://rancher.com/docs/rancher/v2.6/en/installation/other-installation-methods/behind-proxy/install-rancher/#install-cert-manager) section as well?

axeal:
@jtravee no_proxy is the correct value for the cert-manager chart, so no update needed there (https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/values.yaml#L180)

jtravee (approved):

PR #3948, opened 2022-03-14
Added info on Harvester node driver ISO images

### LINKED ISSUE #3889: [Rancher2] Document .ISO usage in Harvester node driver in Rancher
**Request Summary:**
Harvester allows `.ISO` images to be uploaded/displayed through harvester's UI. However, when using harvester node driver (through rancher), `.ISO`s do not display (by design) in rancher. 

**Details:**
We should document this so that users know it is a feature, not a bug. This is mostly because, `.iso` usually require additional setup that interferes with a clean deployment (without requiring user intervention) and are not typically used in cloud environments. 

Since this is specific to using harvester as a node driver in rancher, I would expect this to go in rancher docs. However some cross referencing from harvester docs is recommended to ease user experience.

jtravee:
Closes #3889


nunix (approved):

nunix:
I approved based on the text on Rancher docs as it seems it will not be implemented at all. However, if something is planned to show the `.ISO` in Rancher UI, then maybe we should add a note about the support coming in a later version?

slickwarren (approved):

PR #3973, opened 2022-03-23
Updating cert-manager version for 2.6 docs

### LINKED ISSUE #3921: [Rancher2] Update docs with latest cert-manager in Rancher server helm chart
Docs issue to accompany [cert-manager validation issue in rancher/rancher](https://github.com/rancher/rancher/issues/36693). 

Cert-manager 1.5.1 is EOL and 1.6.x will be EOL 30 March 2022. Update docs with new version once validation is completed.

jtravee:
Closes #3921

- Updated 1.5.x with 1.7.x versions in commands. Added caveat that 1.6.x is compatible but since it is EOL soon, users should be aware.


catherineluse (approved):
LGTM

content/rancher/v2.6/en/installation/install-rancher-on-k8s/upgrades/_index.md
```diff
@@ -150,6 +150,8 @@ helm upgrade rancher rancher-<CHART_REPO>/rancher \
 
 If you are currently running the cert-manager whose version is older than v0.11, and want to upgrade both Rancher and cert-manager to a newer version, then you need to reinstall both Rancher and cert-manager due to the API change in cert-manager v0.11.
```
MbolotSuse, line 151: One thing that I would recommend here - I would recommend changing this sentence to something that reflects that this is also true for 1.5 -> 1.6 Something like this:

```
If you are currently running the cert-manager whose version is 1.5 or below, and want to upgrade both Rancher and cert-manager to a new version (1.6+ in the case of cert-manager), then you need to re-install both Rancher and cert-manager due to the API change in cert-manager 1.6.
```
We may also want to keep the current warning in place, which would likely include another sentence as well:
```
This will also be necessary if you are upgrading from a version of cert manager below 0.11 to a version of cert-manager above 0.11
```

MbolotSuse (changes requested):
Got one addition that might be helpful.

MbolotSuse (commented):
General Comment- 
We also have a page (Upgrading-cert-manager in `installation/resources/upgrading-cert-manager`) which deals with cert-manager upgrades. Do we want to revisit this page in light of this new information?

content/rancher/v2.6/en/installation/install-rancher-on-k8s/upgrades/_index.md
```diff
@@ -148,7 +148,9 @@ helm upgrade rancher rancher-<CHART_REPO>/rancher \
 
 ### Option B: Reinstalling Rancher and cert-manager
 
-If you are currently running the cert-manager whose version is older than v0.11, and want to upgrade both Rancher and cert-manager to a newer version, then you need to reinstall both Rancher and cert-manager due to the API change in cert-manager v0.11. 
+If you are currently running the cert-manager whose version is 1.5 or below, and want to upgrade both Rancher and cert-manager to a new version (1.6+ in the case of cert-manager), then you need to re-install both Rancher and cert-manager due to the API change in cert-manager 1.6. This will also be necessary if you are upgrading from a version of cert manager below 0.11 to a version of cert-manager above 0.11.
+
+>**Note:** New in v2.6.4, cert-manager versions 1.6.2 and 1.7.1 are compatible. We recommend v1.7.x because v 1.6.x will reach end-of-life on March 30, 2022.
```
MbolotSuse, line 153: One small note here: 

I think one of the concerns was that cert-manager does not have a documented route from 1.5-1.7 without an intermediate 1.6 step. It might be good to note that if you are below 1.5 and want to go to 1.6, you should follow the cert-manager docs and do a 1.5-> 1.6 and 1.6 -> 1.7 upgrade. We don't necessarily have to state this anywhere since we don't explicitly endorse this path, but it might be a good note. We could also link t our docs on upgrading cert-manager which link to their docs (although we aren't explicit there either). I don't think that we necessarily need to explicitly state this, but might be good. Let me know what your thoughts are.

jtravee: Revised.

MbolotSuse (approved):

PR #3987, opened 2022-03-28
Updating vSphere CSI and CPI information.

phillipsj:
Currently, what is listed in the docs doesn't match the upstream support matrix for those projects. This updates it to be aligned with that.


sirredbeard (approved):
LGTM

PR #3994, opened 2022-03-30
Make amendments to the installation netowrking requirements

### LINKED ISSUE #3991: [Rancher2] Note for Ember UI being unreachable with X-Frame-Option header
**Request Summary:**

Based on a conversation on the [Rancher forums](https://forums.rancher.com/t/blank-pages-in-rancher-ui-after-upgrade-to-2-6-3/37332), if a server containing Rancher has the `X-Frame-Options=DENY` header the Ember UI will be completely unreachable as it is an embedded `iframe` within the new Dashboard.

I think we should have a note/section explaining this to users who may be upgrading from the legacy UI.

divya-mohan0209:
Fixes #3991

### DIFF
--- a/content/rancher/v2.6/en/installation/requirements/_index.md
+++ b/content/rancher/v2.6/en/installation/requirements/_index.md
@@ -170,7 +170,7 @@ Rancher performance depends on etcd in the cluster performance. To ensure optima
 
 This section describes the networking requirements for the node(s) where the Rancher server is installed.
 
-> If a server containing Rancher has the `X-Frame-Options=DENY` header, the Ember UI will be completely unreachable on upgrading from the legacy UI. This is because any screen written for the legacy UI is rendered as an embedded iframe within the new Dashboard. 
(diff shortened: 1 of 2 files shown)


jordojordo (changes requested):
One small change from what I wrote in the issue, I think the property should be written as: 
`X-Frame-Options: DENY` 
[For reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options#syntax)
Otherwise LGTM

jordojordo (approved):

catherineluse (changes requested):
I think most Rancher users wouldn't know what the Ember UI is because it's in implementation detail. Can we say something like:

> If a server containing Rancher has the `X-Frame-Options: DENY` header, some pages in the new Rancher UI will not be able to render after upgrading from the legacy UI. This is because some legacy pages are embedded as iFrames in the new UI.

divya-mohan0209:
> I think most Rancher users wouldn't know what the Ember UI is because it's in implementation detail. Can we say something like:
> 
> > If a server containing Rancher has the `X-Frame-Options: DENY` header, some pages in the new Rancher UI will not be able to render after upgrading from the legacy UI. This is because some legacy pages are embedded as iFrames in the new UI.

Addressed in the latest commit.

catherineluse (approved):

PR #4004, opened 2022-04-04
Revert "Staging to master" due to rendering problems

catherineluse:
Reverts rancher/docs#4002


catherineluse:
The root cause of the rendering problems was that this Permalink change was supposed to be taken out of the staging branch. https://github.com/rancher/docs/pull/4005

Now that the Permalink change has been reverted in staging and master, I believe it should be safe to un-revert this larger staging-to-master PR. To make sure that's safe I'm doing a local dry run.

catherineluse:
In my local environment I tested what would happen if we merged staging to master (and didn't revert it) and also reverted the Permalink change. So in my test, the git history looks like this:

```
419d6c2cd9217f42241b8346cc062c0e29b4e29e (HEAD -> master) Revert "Use the current hostname for baseUrl"
ddc72593542f5248d1db210d8f5a6f5947f132c6 Merge pull request #4002 from rancher/staging
1fd1455e5920d11cc1573de7566e70f1d0542504 Merge branch 'master' into staging
```

And in that case the docs build successfully with no rendering problems. So I think it's safe to revert the revert.

Issue #4064, opened 2022-05-02
[Rancher2] Update ingress-nginx in hosted cloud provider installation tutorials


bashofmann:
**Request Summary:**
Update ingress-nginx in hosted cloud provider installation tutorials
* https://rancher.com/docs/rancher/v2.6/en/installation/install-rancher-on-k8s/amazon-eks/
* https://rancher.com/docs/rancher/v2.6/en/installation/install-rancher-on-k8s/aks/
* https://rancher.com/docs/rancher/v2.6/en/installation/install-rancher-on-k8s/gke/

**Details:**
The ingress-nginx version that is referenced is already quite old and contains known security vulnerabilities. 

Additionally, newer versions of the upstream ingress-nginx are configured by default to not watch ingresses without an ingress class (https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml#L75). This makes the Rancher installation not work, because the Rancher Helm chart does not set an ingress class, which can lead to confusion (see https://github.com/rancher/rancher/issues/37193). The tutorial should be updated to configure ingress-nginx to also watchin ingress without an ingressClass.

bashofmann:
Done with https://github.com/rancher/docs/pull/4227

PR #4065, opened 2022-05-02
[Rancher] Update ingress-nginx in the hosted kubernetes install tutorials

bashofmann:
The newer ingress-nginx fixes known security vulnerabilities and is fully compatible with Rancher. Since the Rancher Helm chart does not set an ingressClass on the Rancher ingress, `controller.watchIngressWithoutClass` must be set to `true` (default value in newer ingress-nginx versions is `false`).

Fixes https://github.com/rancher/docs/issues/4064. See also https://github.com/rancher/rancher/issues/37193.

SURE-4878

### DIFF
--- a/content/rancher/v2.6/en/installation/resources/k8s-tutorials/aks/_index.md
+++ b/content/rancher/v2.6/en/installation/resources/k8s-tutorials/aks/_index.md
@@ -81,7 +81,8 @@ helm upgrade --install \
   ingress-nginx ingress-nginx/ingress-nginx \
   --namespace ingress-nginx \
   --set controller.service.type=LoadBalancer \
-  --version 3.12.0 \
+  --set controller.watchIngressWithoutClass=true \
+  --version 4.1.0 \
   --create-namespace
 ```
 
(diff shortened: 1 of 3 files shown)


snasovich:
@bashofmann , thank you for submitting these changes. I'm wondering where does "and is fully compatible with Rancher" portion come from? I'm not sure how we approach testing for versions of ingress-nginx for AKS/GKE/EKS.
cc: @sowmyav27 @slickwarren

bashofmann:
@snasovich RKE and RKE2 are already bundling newer versions of the ingress controller as well.

snasovich:
@kinarashah , could you please review this?

What caught my eye is that we're setting `watchIngressWithoutClass` to `true` by default in the template used by k8s <1.22 (https://github.com/rancher/kontainer-driver-metadata/blob/1d2c195db09f7e865309594a1e2ae188c369177b/rke/templates/nginx-v1.2.0-rancher1.go#L380) but for the template used by k8s >=1.22.9 we do not (https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/rke/templates/nginx-v1.2.0-rancher2.go). Though we do set annotation `ingressclass.kubernetes.io/is-default-class: "true"` in both cases - not sure if it takes care of the same.

snasovich:
@bashofmann , apologies for no following up on this in a timely fashion.
In my comment, I was simply pointing out that we have stopped setting this for RKE1 downstream clusters, so was concerned there may have been a good reason we do this.
Asked @kinarashah to chime in to be sure as she's the Ingress expert. I'll have her take another look.

btat:
@kinarashah @snasovich wondering if you've had a chance to look over this yet? We've had folks on Slack run into/ask about this recently.

bashofmann:
Done with https://github.com/rancher/docs/pull/4227.

PR #4073, opened 2022-05-04
Update _index.md

DanielCerejo:
correct Example docker push mycustomreg:5000 should be mycustomreg.com:5000


DanielCerejo:
i thing cold be more confusions/erros in other parts off documentation  with  mycustomreg:5000 and not mycustomreg.com:5000.

brandond (approved):

PR #4076, opened 2022-05-06
Show example using installer

cawoodm:
The current doc assumes that k3s is already installed.
As most users use the installer to initiate a cluster it is confusing not having an example with the installer.

### DIFF
--- a/content/k3s/latest/en/installation/ha-embedded/_index.md
+++ b/content/k3s/latest/en/installation/ha-embedded/_index.md
@@ -15,7 +15,6 @@ To run K3s in this mode, you must have an odd number of server nodes. We recomme
 To get started, first launch a server node with the `cluster-init` flag to enable clustering and a token that will be used as a shared secret to join additional servers to the cluster.
 ```
 K3S_TOKEN=SECRET
-export INSTALL_K3S_VERSION=v1.23.6+k3s1
 curl -sfL https://get.k3s.io | sh -s - server --cluster-init
 ```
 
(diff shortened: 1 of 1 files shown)


content/k3s/latest/en/installation/ha-embedded/_index.md
```diff
@@ -14,7 +14,9 @@ To run K3s in this mode, you must have an odd number of server nodes. We recomme
 
 To get started, first launch a server node with the `cluster-init` flag to enable clustering and a token that will be used as a shared secret to join additional servers to the cluster.
 ```
-K3S_TOKEN=SECRET k3s server --cluster-init
+K3S_TOKEN=SECRET
+export INSTALL_K3S_VERSION=v1.23.6+k3s1
```
dereknola, line 18:
```suggestion
```
Including the version is unnecessary and confusing in this example.

dereknola (changes requested):
This is the correct way to use `--cluster-int` from the installer.

content/k3s/latest/en/installation/ha-embedded/_index.md
cawoodm: > This is the correct way to use `--cluster-int` from the installer.
Are you saying it's correct that `--cluster-init` is passed to the installer and embedded in the systemd config? So k3s is run with `--cluster-init` on one of my nodes each time that node starts??

btat:
Hi @dereknola , does this look good to merge now?

dereknola (approved):

PR #4086, opened 2022-05-10
Update RKE2 config ref YAML options

btat:


content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -6,13 +6,36 @@ weight: 5
...
+
+# Editing Clusters with YAML
+
+Instead of using the Rancher UI to choose Kubernetes options for the cluster, advanced users can create an RKE2 config file. Using a config file allows you to set any of the options available in an RKE2 installation by specifying them in YAML.
+
+To edit an RKE2 config file directly from the Rancher UI,
+
+1. Click **☰ > Cluster Management**.
+1. Go to the cluster you want to configure and click **⋮ > Edit as YAML**.
+1. Edit the RKE options under the `rancher_kubernetes_engine_config` directive.
```
sowmyav27, line 37: For RKE2 clusters it is under `rkeConfig`

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -6,13 +6,36 @@ weight: 5
...
+
 # Configuration Options in the Rancher UI
```
sowmyav27, line 39: https://github.com/rancher/docs/blob/735b4ccb7c663395be8e3554fe668033d74e929f/content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md#selinux

```
#### SELinux
Option to enable or disable [SELinux](https://docs.rke2.io/security/selinux) support.
```
This option is NOT available from Rancher UI.

sowmyav27 (dismissed):

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
btat: Updated

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
btat: Removed.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -161,10 +184,110 @@ Option to remove all pods from the node prior to upgrading.
...
 Option to set kubelet options for different nodes. For available options, refer to the [Kubernetes documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).
 
-# Cluster Config File
-
-Instead of using the Rancher UI forms to choose Kubernetes options for the cluster, advanced users can create an RKE2 config file. Using a config file allows you to set any of the [options](https://docs.rke2.io/install/install_options/install_options) available in an RKE2 installation.
-
-To edit an RKE2 config file directly from the Rancher UI, click **Edit as YAML**.
-
-
+#### Kubernetes Node Labels
```
snasovich, line 183: "Kubernetes Node Labels" is under "Advanced" on each Machine Pool while this "Advanced" section in the documentation is for "Cluster Configuration" part.

In fact, it doesn't look like this doc page explains Machine Pools settings at all. Is this something that's planned to be covered as part of another PR? If not, we want to add explanation of "Machine Pools" section and settings available there (note that part of that UI depends on what infrastructure provider is selected, so that's a good opportunity to link to parent page of where we specify details for DO, AWS, Azure etc.).

While at it, "Labels & Annotations" section from "Cluster Configuration" doesn't seem to be covered at all (the reason I've noticed is that I was looking where Labels should be populated since we're talking about labels here).

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -161,10 +184,110 @@ Option to remove all pods from the node prior to upgrading.
...
+        deleteEmptyDirData: true
+        enabled: true
+        gracePeriod: -1
+        ignoreDaemonSets: true
+        timeout: 120
+```
+
+### chartValues
+
+Option to install charts and specify their values. This option is intended to be used for RKE2/K3s system charts. All charts install using this method are installed in the `kube-system` namespace.
```
snasovich, line 242:
```suggestion
Option to specify the values for the system charts installed by RKE2/k3s.
```
Don't think we have to mention namespace here.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -161,10 +184,110 @@ Option to remove all pods from the node prior to upgrading.
...
+
+Example:
+
+```yaml
+chartValues:
+    chart-name:
+        key: value
+```
+
+This would install the chart with `chart-name` with the values `key = value`.
```
snasovich, line 252: This statement gives the impression that we specify which chart to install which I don't believe is the case. E.g. if no chartValues are specified, RKE2 still installs whatever charts it needs; this section is purely to allow customizing values for the charts RKE2 installs (either by default or based on other settings passed - e.g. `machineGlobalConfig.cni` sets which CNI is to be installed and related Helm chart is installed automatically).

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -161,10 +184,110 @@ Option to remove all pods from the node prior to upgrading.
...
+
+### machineGlobalConfig
+
+The RKE2/K3s configurations are nested under the `machineGlobalConfig` directive. Any configuration change made here will apply to every node. All configuration options available in the [standalone version of RKE2](https://docs.rke2.io/install/install_options/server_config/) can also be applied here.
```
snasovich, line 256: Related to the other comment, aren't we limited to options specified on `serverArgs` and `agentArgs` as set in KDM for specific RKE2/k3s version, e.g. https://github.com/rancher/kontainer-driver-metadata/blob/c974bcd1461419789c59066038fc69dce34168d0/channels-rke2.yaml#L21

If that's so, it make it less of "all options" but more of "select options". It will be rather hard to explain what those select options are though.

@thedadams , am I correct here?

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -161,10 +184,110 @@ Option to remove all pods from the node prior to upgrading.
...
+
+Example:
+
+```yaml
+machineGlobalConfig:
+    etcd-arg:
+        key1: value1
+        key2: value2
```
snasovich, line 264: Per Donnie's example it should be a little different:

```suggestion
machineGlobalConfig:
    etcd-arg:
        - key1=value1
        - key2=value2
```

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -161,10 +184,110 @@ Option to remove all pods from the node prior to upgrading.
...
+Example:
+
+```yaml
+machineSelectorConfig
+  - config:
+      config-key: config-value
+    machineLabelSelector:
+      matchExpressions:
+        - key: example-key
+          operator: string # Like Exists, In, NotIn
```
snasovich, line 282: I believe the complete list of operators is "In, NotIn, Exists, and DoesNotExist" so instead of giving 3/4 under "like" we should just list all 4. Plus "Like" almost looks like an option for the operator here.
@thedadams , is this the complete list?

snasovich (changes requested):
Added some suggestions.
Would appreciate for @thedadams to review my comments as I'm not exactly solid on their correctness.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
snasovich: Good catch @sowmyav27 ! https://github.com/rancher/dashboard/issues/5025

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -6,13 +6,36 @@ weight: 5
...
+- [Editing Clusters with YAML](#editing-clusters-with-yaml)
+- [Configuration Options in the Rancher UI](#configuration-options-in-the-rancher-ui)
+- [Cluster Config File Reference](#cluster-config-file-reference)
+
 # Overview
 
 You can configure the Kubernetes options in one of the two following ways:
 
 - [Rancher UI](#configuration-options-in-the-rancher-ui): Use the Rancher UI to select options that are commonly customized when setting up a Kubernetes cluster.
 - [Cluster Config File](#cluster-config-file): Instead of using the Rancher UI to choose Kubernetes options for the cluster, advanced users can create an RKE2 config file. Using a config file allows you to set any of the [options](https://docs.rke2.io/install/install_options/install_options) available in an RKE2 installation.
```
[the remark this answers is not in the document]
thedadams: Yes, you're right, Sergey. Technically, a user can specify any of those options (or whatever they want) and Rancher will filter out the unsupported options based on what is in KDM. I'm not sure how we want to document which parameters are configurable from Rancher.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
thedadams: The "Labels & Annotations" section in the cluster configuration refers to labels and annotations that go on the objects in Rancher and not related to the downstream cluster configuration at all.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
thedadams: Technically speaking, I believe it does mean that Rancher will ask RKE2 to install the chart. The issue is whether or not RKE2 knows about the chart. If RKE2 doesn't know about the chart, then Rancher asking it to install it won't work.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
thedadams: The flow here is:
1. Everything under `machineGlobalConfig` gets applied to all node configurations
2. Before that configuration is "saved" and applied to the node, Rancher will filter out all configuration that is unsupported.
3. The configuration is saved and sent to the node.

Not saying we should include this in the documentation because it is an implementation detail, but that is how it works.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
thedadams: According to [this](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement), the complete list is already include.

Perhaps removing the "Like" is a good idea.

thedadams: The `matchExpressions` section is more like a [label set](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement)

thedadams (changes requested):

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
snasovich: Thank you for the clarification here. If we make the suggested change few lines above, I think this can be left as is.
On a side note, I'm wondering if there is a way to reference a comprehensive list of charts coming from RKE2?
I couldn't find it documented in RKE2 docs.
@cwayne18 @brandond @briandowns

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
snasovich: Similar to the other comment, I think a good first step is to remove word "all":
```suggestion
The RKE2/K3s configurations are nested under the `machineGlobalConfig` directive. Any configuration change made here will apply to every node. The configuration options available in the [standalone version of RKE2](https://docs.rke2.io/install/install_options/server_config/) can be applied here.
```

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
snasovich: I've noticed this in the https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements
> Valid operators include In, NotIn, Exists, and DoesNotExist.

Is "DoesNotExist" not supported here?

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
btat: `Kubernetes Node Labels` moved under the `Machine Pools` area now.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
brandond: I don't believe this does enable the chart. It just creates a HelmChartConfig for the chart with the matching name, which will provide values to that chart IF it is enabled. I would love to have this confirmed though.

The RKE2 bundled charts live in the [rke2-charts repo](https://github.com/rancher/rke2-charts/tree/main/charts), and the versions packaged in a given release can be found in the [Dockerfile](https://github.com/rancher/rke2/blob/v1.23.6+rke2r2/Dockerfile#L107-L118). Some of these, such as the cloud-provider and CNI charts, only have one active at a time. We don't document this anywhere else, at the moment.

brandond (dismissed):

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
thedadams: The documentation link you supplied seems to imply that `DoesNotExist` is supported.

And from the `apimachinery` code: `// Valid operators are In, NotIn, Exists and DoesNotExist.`

So I think the documentation should be updated to reflect this.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
btat: Would the consensus be to remove the line for now and add back any additional info once it's been confirmed?

@snasovich @thedadams @brandond

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
btat: List of operators updated.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
snasovich: I agree, let's just remove this line - the documentation is fine without it.

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -6,17 +6,83 @@ weight: 5
...
+- [Configuration Options in the Rancher UI](#configuration-options-in-the-rancher-ui)
+- [Cluster Config File Reference](#cluster-config-file-reference)
+
 # Overview
 
 You can configure the Kubernetes options in one of the two following ways:
 
 - [Rancher UI](#configuration-options-in-the-rancher-ui): Use the Rancher UI to select options that are commonly customized when setting up a Kubernetes cluster.
-- [Cluster Config File](#cluster-config-file): Instead of using the Rancher UI to choose Kubernetes options for the cluster, advanced users can create an RKE2 config file. Using a config file allows you to set any of the [options](https://docs.rke2.io/install/install_options/install_options) available in an RKE2 installation.
+- [Cluster Config File](#cluster-config-file): Instead of using the Rancher UI to choose Kubernetes options for the cluster, advanced users can create an RKE2 config file. Using a config file allows you to set all the [options](https://docs.rke2.io/install/install_options/install_options) available for an RKE2 installation.
```
snasovich, line 20:
```suggestion
- [Cluster Config File](#cluster-config-file): Instead of using the Rancher UI to choose Kubernetes options for the cluster, advanced users can create an RKE2 config file. Using a config file allows you to set the [options](https://docs.rke2.io/install/install_options/install_options) available for an RKE2 installation.
```

Looks like we're going back and forth on this. Per the conversation in https://github.com/rancher/docs/pull/4086#discussion_r870836269 I don't believe stating "all" is technically correct. I see it was re-added based on the suggestion from @brandond , so maybe it's correct. Still, removing "all" makes it "safer".

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -6,17 +6,83 @@ weight: 5
...
+
+The duration nodes can be unreachable before they are automatically deleted and replaced.
+
+#### Drain Before Delete
+
+Enables draining nodes by evicting all pods before the node is deleted.
+
+#### Kubernetes Node Labels
+
+Add [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/), to nodes to help with organization and object selection.
```
snasovich, line 76:
```suggestion
Add [labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) to nodes to help with organization and object selection.
```
Nit - remove comma, I don't believe it's necessary here

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
```diff
@@ -181,8 +243,106 @@ Option to remove all pods from the node prior to upgrading.
...
 Option to set kubelet options for different nodes. For available options, refer to the [Kubernetes documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).
 
-# Cluster Config File
-
-Instead of using the Rancher UI forms to choose Kubernetes options for the cluster, advanced users can create an RKE2 config file. Using a config file allows you to set any of the [options](https://docs.rke2.io/install/install_options/install_options) available in an RKE2 installation.
-
-To edit an RKE2 config file directly from the Rancher UI, click **Edit as YAML**.
+# Cluster Config File Reference
+
+Instead of using the Rancher UI to choose Kubernetes options for the cluster, advanced users can create a config file. Using a config file allows you to set any of the [options available](https://docs.rke2.io/install/install_options/server_config/) in an RKE2 installation, including those already listed in [Configuration Options in the Rancher UI](#configuration-options-in-the-rancher-ui), as well as Rancher-specific parameters.
```
snasovich, line 248:
```suggestion
Instead of using the Rancher UI to choose Kubernetes options for the cluster, advanced users can create a config file. Using a config file allows you to set the [options available](https://docs.rke2.io/install/install_options/server_config/) in an RKE2 installation, including those already listed in [Configuration Options in the Rancher UI](#configuration-options-in-the-rancher-ui), as well as Rancher-specific parameters.
```
Same as in previous comment suggesting to remove "any of"

snasovich (changes requested):
Some minor updates requested, otherwise LGTM

content/rancher/v2.6/en/cluster-admin/editing-clusters/rke2-config-reference/_index.md
brandond: It doesn't sound right if we remove a quantifier: "Using a config file allows you to set the options available for an RKE2 installation." Maybe we can say something like:

```suggestion
- [Cluster Config File](#cluster-config-file): Instead of using the Rancher UI to choose Kubernetes options for the cluster, advanced users can create an RKE2 config file. Using a config file allows you to set many additional the [options](https://docs.rke2.io/install/install_options/install_options) available for an RKE2 installation.
```

Do we have any documentation, internal or otherwise, on what options **cannot** be configured this way?

btat: @snasovich would you be able to answer these so we can move this PR along?

snasovich: @btat , thank you for following up on this. 
I personally like the suggestion by @brandond.
As to what cannot be configured this way, I believe it's limited to agent and server args as set in KDM: https://github.com/rancher/kontainer-driver-metadata/blob/dev-v2.6/channels-rke2.yaml

However, there are some caveats, for example around parameters that reference paths (e.g. `audit-policy-file`) - the path can be specified, but I guess we should allow to provide actual file contents and have provisioning place that file on the nodes and set the `audit-policy-file` to the path of this file. This is currently part of https://github.com/rancher/rancher/issues/32784.
@Oats87 , please correct/expand as necessary.

PR #4115, opened 2022-05-21
[TYPO]Update _index.md

FingerLiu:


jtravee:
Thank you for your suggestion. In this particular scenario, the word `elect` is more fitting than `select` both in grammatical structure and intent. It's actually not a typo, but rather the word `elect` is being used in place of `decide`, i.e., if you decide (elect) to use Rancher..., you must add cert-manager...". Closing as no change is needed.

PR #4117, opened 2022-05-23
Update _index.md to correct misleading / incorrect statements regarding NeuVector

skywalke34:
Removing incorrect statements regarding what NeuVector does or doesn't support.
NeuVector DOES support air-gapped environments.
NeuVector DOES support pre-hardened containers - Why would someone say it doesn't???

### DIFF
--- a/content/rancher/v2.6/en/neuvector-integration/_index.md
+++ b/content/rancher/v2.6/en/neuvector-integration/_index.md
@@ -104,12 +104,6 @@ Below are the minimum recommended computing resources for the NeuVector chart in
 
 * NeuVector is not supported on a Windows cluster.
 
-* NeuVector installation is not supported on hardened clusters.
-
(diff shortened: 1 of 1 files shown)


sowmyav27:
Neuvector is working on 2.6.5. However, the only caveat is that the image has to be updated regularly/daily in the registry. 

And these are something Rancher QA has not tested and was deemed out of scope for 2.6.5.
```
* NeuVector installation is not supported on hardened clusters.
* NeuVector installation is not supported on SELinux clusters.
```
@selvamt94 Can you please comment if these two usecases are supported? ^^

selvamt94:
In phase 1 NeuVector integration, Rancher Dev just wanted to support basic SSO because other support needs more time to test and fine-tune the feature chart. All this support will be available in phase2 next release.

jtravee:
Closing as there is no need for a docs update at this time.

PR #4145, opened 2022-06-09
cert-manager url change

sterni1971:
kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.7.1/cert-manager.crds.yam
error: unable to read URL "https://github.com/jetstack/cert-manager/releases/download/v1.7.1/cert-manager.crds.yam", server reported 404 Not Found, status code=404


btat (approved):

btat:
Thanks for the contribution @sterni1971. 

As for the error you quoted, it looks like you left off the `l` at the end of `yaml`. The current URL, while not up-to-date, should work due to redirects.

PR #4147, opened 2022-06-09
Updating deprecation page

jtravee:
Updating page to remove 2.5x verbiage/added patch-specific version links. At this time, there isn't a more in-depth deprecation policy doc. Slack conversation:

So our docs point to [information about deprecation here](https://rancher.com/docs/rancher/v2.6/en/faq/deprecated-features-25x/),  there are still some customers who have questions about what is considered "legacy" after updating to 2.6. Was there a more in-depth deprecation policy document?


btat (approved):
lgtm

PR #4165, opened 2022-06-22
Called out lack of additional logic around default Kubernetes service extra_args

### LINKED ISSUE #4142: [RKE] Call out lack of additional logic around default kubernetes service extra_args
**Summary:**
Via the extra_args settings [https://rancher.com/docs/rke/latest/en/config-options/services/services-extras/#extra-args] on the default Kubernetes services in an RKE cluster configuration YAML, it is possible for users to define additional arguments on these cluster components. RKE adds these arguments to the components without any additional logic, i.e. in the case that a particular argument requires another action when changed, such as the kubelet cpu-manager-policy arguments, which requires removal of the cpu_manager_state file if it is changed, this action is not performed automatically by RKE and will require manual user intervention to remove the file from the cluster nodes (https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#changing-the-cpu-manager-policy). The request here is to call-out this behaviour in which the extra_args are added to services without additional logic, so any changes requiring additional steps would need to be handled by the user/may not be compatible with RKE.

jtravee:
Closes #4142 

Reference: [`extra_args` section](https://rancher.com/docs/rke/latest/en/config-options/services/services-extras/#extra-args)

### DIFF
--- a/content/rke/latest/en/config-options/services/services-extras/_index.md
+++ b/content/rke/latest/en/config-options/services/services-extras/_index.md
@@ -26,9 +26,9 @@ services:
         cluster-name: "mycluster"
 ```
 
-#### Defining Additional Arguments Manually
+#### Lack of Additional Logic in Additional Arguments
 
-- It is possible for users to add additional arguments on the default Kubernetes services in the RKE cluster configuation YAML. RKE will add these arguments to the components **without** any additional logic, however. 
+- RKE will add additional arguments to the Kubernetes cluster components **without** any additional logic. 
 
 - Because no additional logic is added when `extra_args` are added to services, users **must** perform any additional steps necessary manually to ensure these changes are compatible with RKE.
 
--- a/content/rke/latest/en/config-options/services/services-extras/_index.md
+++ b/content/rke/latest/en/config-options/services/services-extras/_index.md
@@ -3,7 +3,7 @@ title: Extra Args, Extra Binds, and Extra Environment Variables
 weight: 231
 ---
 
-RKE supports additional service arguments, volume binds and environment variables.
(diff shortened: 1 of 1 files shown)


content/rke/latest/en/config-options/services/services-extras/_index.md
```diff
@@ -26,6 +26,14 @@ services:
         cluster-name: "mycluster"
 ```
 
+#### Defining Additional Arguments Manually
```
btat, line 29: The header should be about (the lack of) additional logic.

content/rke/latest/en/config-options/services/services-extras/_index.md
```diff
@@ -26,6 +26,14 @@ services:
...
+
+- It is possible for users to add additional arguments on the default Kubernetes services in the RKE cluster configuation YAML. RKE will add these arguments to the components **without** any additional logic, however.
```
btat, line 31: The first sentence is not needed since it's describing what extra_args does, which is already done in the intro sentence of the `### Extra Args` parent section.

btat (changes requested):

content/rke/latest/en/config-options/services/services-extras/_index.md
jtravee: Rephrased header.

content/rke/latest/en/config-options/services/services-extras/_index.md
jtravee: Removed.

axeal (approved):

btat (approved):

PR #4172, opened 2022-06-29
Update quickstart links

outscale-mdr:
[rancher/quickstart](https://github.com/rancher/quickstart) repository has recently changed its structure, some links were not updated.


jtravee:
@outscale-mdr, thank you for your contribution! Would you please also update the 2.5.x links as well?

jtravee:
> Hi @jtravee, of course, here you go.
> 
> For the version before v2.5, do you still maintain it ? (links are broken too)

Thanks! We've been maintaining them mainly for certain items, primarily critical fixes. In this case, since @btat  had already updated the commands for 2.5, it made sense to also update the links. Overall, we support 2.5 docs as long as 2.5 is in support, which keeps being extended essentially!

PR #4175, opened 2022-06-30
Minor correction of k3s HA embedded DB notes

nlucansk:
Adding a small update to the notes.

image

### DIFF
--- a/content/k3s/latest/en/installation/ha-embedded/_index.md
+++ b/content/k3s/latest/en/installation/ha-embedded/_index.md
@@ -14,15 +14,25 @@ To run K3s in this mode, you must have an odd number of server nodes. We recomme
 
 To get started, first launch a server node with the `cluster-init` flag to enable clustering and a token that will be used as a shared secret to join additional servers to the cluster.
 ```
-curl -sfL https://get.k3s.io | K3S_TOKEN=SECRET sh -s - server --cluster-init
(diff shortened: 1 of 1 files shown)


brandond:
These are supposed to be commands that users can copy-paste directly into their shell with as few modifications as possible. If you put a $ at the beginning of the line then users have to be sure not to copy that into their shell or they will get errors. What is the intent in adding that?

nlucansk:
> These are supposed to be commands that users can copy-paste directly into their shell with as few modifications as possible. If you put a $ at the beginning of the line then users have to be sure not to copy that into their shell or they will get errors. What is the intent in adding that?

I was trying to sync the style of "double check" section.
But sure I will add a copy / paste modification in a few moments

brandond:
Yeah, we could probably be more consistent about showing where we're providing sample output of a command (as in your `kubectl get nodes` block) and where we we are providing script blocks that we intend people to copy-paste into their shell.

nlucansk:
> Yeah, we could probably be more consistent about showing where we're providing sample output of a command (as in your `kubectl get nodes` block) and where we we are providing commands that we intend people to copy-paste into their shell. Its possible we could use different formatters for the markdown blocks? on GH at least `bash` can be used for commands, while `console` is good for example CLI input + output.

I am 🤔  about collapsible output.
It can be also implemented for the other docs. Experienced user can just roll across the docs and do the thing.
Newcomer can read the docs line by line and use the outputs to compare results.

image

brandond (approved):

PR #4176, opened 2022-06-30
Adding note about disabling firewalld in RKE1

### LINKED ISSUE #4174: Change RKE documentation to recommend disabling firewalld, as we do elsewhere.
**Summary:**
In our Rancher guide, as well as the RKE2 guide we recommend disabling firewalld altogether. 
the RKE guide is the exception where we have some OS specific recommendations, yet these often turn out not to be enough when features like Project Network Isolation or PNI are in use. 
Its easier just to recommend disabling firewalld as we do elsewhere and be consistent. 

**Details:**
Example:
https://rancher.com/docs/rke/latest/en/os/#red-hat-enterprise-linux-rhel-oracle-linux-ol-centos

The steps described here are not enough, to allow PNI to work if firewalld is enabled.

jtravee:
Closes #4174

### DIFF
--- a/content/rke/latest/en/os/_index.md
+++ b/content/rke/latest/en/os/_index.md
@@ -2,6 +2,9 @@
 title: Requirements
 weight: 5
 ---
+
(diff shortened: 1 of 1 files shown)


vincebrannon (approved):
Looks good.

PR #4183, opened 2022-07-04
Include an alternative to the sh command:

osg:
When I ran the command as-is, I did not receive an error message that could help me fix the issue. My output looked as follows:

$ curl -sfL https://get.k3s.io | sh -
[INFO]  Finding release for channel stable
[INFO]  Using stable as release
[INFO]  Downloading hash https://github.com/k3s-io/k3s/releases/download/stable/sha256sum-amd64.txt
$ k3s kubectl get node

Command 'k3s' not found, did you mean:

  command 'k9s' from snap k9s (0.7.12)
  command 'k3b' from deb k3b (19.12.3-0ubuntu1)
  command 'kds' from deb kylin-display-switch (1.0.4-1)

See 'snap info <snapname>' for additional versions.

### DIFF
--- a/content/k3s/latest/en/quick-start/_index.md
+++ b/content/k3s/latest/en/quick-start/_index.md
@@ -16,6 +16,8 @@ K3s provides an installation script that is a convenient way to install it as a 
 curl -sfL https://get.k3s.io | sh -
 ```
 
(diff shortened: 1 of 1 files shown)


dereknola:
This is currently a bug with the stable channel. No need to change the docs.

PR #4188, opened 2022-07-09
Add new docs landing page

catherineluse:
Screen Shot 2022-09-01 at 11 10 17 AM

when you scroll down:
Screen Shot 2022-07-21 at 1 01 23 AM


divya-mohan0209:
This looks good to me, tbh! However, I agree that we need to have graphics eventually if not right at the beginning.

btat:
@catherineluse is https://github.com/rancher/docs/pull/4188/commits/4396009f465e85fcacf58b8834e19053da7ed7e0 the last bit of this PR or are there plans to add more updates?

Also, it looks like the .docusaurus directory and markdown files that were used for testing got checked in, could you update that as well?

catherineluse:
@btat Before this PR is merged we probably want to make at least one more change to make the background lighter or more transparent to make the text easier to read.

Before it is merged I'll also revert the deletion of every content file. I didn't add any Markdown files for testing - there were just a few remaining after I deleted almost everything to make it compile. When I undelete those files, the repo will return to a broken state.

The version of this PR to get merged to master will only update 3 files - the index file for the home page and two CSS files.

It should be noted that even after the PR is merged, the landing page still can't be considered done because the search doesn't work and the links will need to be updated to point to the new docs.

catherineluse:
Closing per feedback from management.

PR #4202, opened 2022-07-15
Add blurb about enabling encryption on install

gmemstr:
It took me a little bit to connect the dots myself, but it's worth adding a little bit about adding the secrets encryption flag during install (since it can't be enabled later).


brandond:
cc @dereknola for docs reorg

btat:
Addressed in https://github.com/k3s-io/docs/pull/13

PR #4208, opened 2022-07-16
missing v in -version=v2.x.x

asemen2608:
missing v in -version=v2.6.6 and update to latest version v2.6.6


btat (changes requested):
The `--version` flag specifies the chart version, which does not have a `v`. Rather than adding a `v` to the `helm upgrade` command, could you remove it from the `helm fetch` one instead. The command seems to work with or without the `v`, but we should list things according to Helm's conventions.

```
helm search repo rancher-latest                                                        
NAME                    CHART VERSION   APP VERSION     DESCRIPTION
rancher-latest/rancher  2.6.6           v2.6.6          Install Rancher Server to manage Kubernetes clu...
```

btat:
Moved to our new repo as https://github.com/rancher/rancher-docs/pull/217.

PR #4225, opened 2022-07-27
Update SSH requirements

rahendatri:
add other sshd_config params

### DIFF
--- a/content/rke/latest/en/os/_index.md
+++ b/content/rke/latest/en/os/_index.md
@@ -330,7 +330,7 @@ Your SSH server system-wide configuration file, located at `/etc/ssh/sshd_config
 AllowTcpForwarding yes
 ```
   
-In some cases (e.g. RHEL 8), you also need to set below parameters
+In some cases (e.g., RHEL 8), you also need to set below parameters:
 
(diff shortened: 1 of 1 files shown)


rahendatri:
closes https://github.com/rancher/rke/issues/2907

content/rke/latest/en/os/_index.md
```diff
@@ -329,3 +329,12 @@ Your SSH server system-wide configuration file, located at `/etc/ssh/sshd_config
 ```
 AllowTcpForwarding yes
 ```
+  
+In some cases (e.g. RHEL 8), you also need to set below parameters
```
jtravee, line 333:
```suggestion
In some cases (e.g., RHEL 8), you also need to set below parameters:
```

jtravee (changes requested):

rahendatri:
@kinarashah @jtravee Hello, it seems like this PR is still left unmerged.
The requested change is already taken into account. Would you be able to process this ?
Thanks in advance.

snasovich:
@jtravee @kinarashah , please review this when you have a chance.

PR #4227, opened 2022-07-27
Updated nginx versions and added note on new `-set ingress.ingressClassName=nginx` value

jtravee:
Closes #4221 

- Updated nginx versions and added note on new value for GKE, AKS, and Amazon EKS
- For the K8s versions, we updated GKE, AKS, and EKS with a placeholder in this [PR](https://github.com/rancher/docs/pull/4218)

PR #4263, opened 2022-08-17
Update wording and version for Ingress Helm chart

frank-at-suse:


jtravee:
@frank-at-suse, thanks for your contribution. We can close this in favor of [#4227](https://github.com/rancher/docs/pull/4227) which has updated the nginx versions and [#4218](https://github.com/rancher/docs/pull/4218) which addresses the K8s version for GKE. Please note that 4227 is in staging and will be merged into master when 2.6.7 releases.

PR #4265, opened 2022-08-18
Add docs for `--egress-selector-mode` flag

mallardduck:
And add `--flannel-ipv6-masq ` to command help output.
NOTE, this is already documented on the "Network Options" page, so doesn't need to be added to a table.


mallardduck (commented):
add missing value placeholder

dereknola:
This has already been added to the new docs site. The old docs are no longer being updated. https://k3s-io.github.io/docs/docs/reference/server-config#networking

PR #4272, opened 2022-08-23
Update docs with the workaround

manuelbuil:

### DIFF
--- a/content/k3s/latest/en/installation/network-options/_index.md
+++ b/content/k3s/latest/en/installation/network-options/_index.md
@@ -87,6 +87,12 @@ Note that you can choose whatever `cluster-cidr` and `service-cidr` value, howev
 
 If you are using a custom cni plugin, i.e. a cni plugin different from flannel, the previous configuration might not be enough to enable dual-stack in the cni plugin. Please check how to enable dual-stack in its documentation and verify if network policies can be enabled.
 
+> **Warning:** Kubernetes 1.24 and 1.25 includes a bug that ignores the node ipv6 addresses if you have a dual-stack environment and you are not using the default NIC for the cluster traffic. To avoid this bug, add the following flag to both k3s servers and agents:
+```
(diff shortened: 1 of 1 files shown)


content/k3s/latest/en/installation/network-options/_index.md
```diff
@@ -87,6 +87,12 @@ Note that you can choose whatever `cluster-cidr` and `service-cidr` value, howev
 
 If you are using a custom cni plugin, i.e. a cni plugin different from flannel, the previous configuration might not be enough to enable dual-stack in the cni plugin. Please check how to enable dual-stack in its documentation and verify if network policies can be enabled.
 
+> **Warning:** Kubernetes 1.24 and 1.25 includes a bug that ignores the node ipv6 addresses if you have a dual-stack environment and you are not using the default NIC for the cluster traffic. To avoid this bug, add the following flag to both k3s servers and agents:
```
jtravee, line 90:
```suggestion
> **Warning:** Kubernetes 1.24 and 1.25 include a bug that ignores the node IPv6 addresses if you have a dual-stack environment and you are not using the default network interface card (NIC) for the cluster traffic. To avoid this bug, add the following flag to both K3s servers and agents:
```

content/k3s/latest/en/installation/network-options/_index.md
```diff
@@ -87,6 +87,12 @@ Note that you can choose whatever `cluster-cidr` and `service-cidr` value, howev
...
+```
+--kubelet-arg "--node-ip=0.0.0.0" # If you want to prioritize IPv6 traffic, use "--node-ip=::" instead of "--node-ip=0.0.0.0"
```
jtravee, line 92:
```suggestion
--kubelet-arg "--node-ip=0.0.0.0" # If you want to prioritize IPv6 traffic, use "--node-ip=::" instead of "--node-ip=0.0.0.0".
```

jtravee (changes requested):

PR #4285, opened 2022-08-30
Adding instructions to use v1.0.1 image

MbolotSuse:
Related to rancher/rancher#38712. The published, first version of the adapter contains a bug that would cause issues if used. Because of this, customers will need to specify a newer version when installing the adapter. This pr updates the install instructions to make use of this new version.

### DIFF
--- a/content/rancher/v2.6/en/installation/cloud-marketplace/aws/adapter-install/_index.md
+++ b/content/rancher/v2.6/en/installation/cloud-marketplace/aws/adapter-install/_index.md
@@ -59,22 +59,26 @@ For the below instructions, replace `$MY_ACC_NUM` with your AWS account number a
 
 > **Note:** If you use shell variables, do not specify quotation marks. For example, MY_ACC_NUM=123456789012 will work, but MY_ACC_NUM="123456789012" will fail.
 
-> **Note:** Accounts using the AWS Marketplace listing for Europe, the Middle East, and Africa (EMEA) will need to specify an additional `--set image.repository=rancher/rancher-csp-adapter-eu` option. To see if your account needs this option when installing the adapter, refer to the usage instructions of the marketplace listing.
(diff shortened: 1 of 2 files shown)


content/rancher/v2.6/en/installation/cloud-marketplace/aws/adapter-install/_index.md
```diff
@@ -65,13 +65,15 @@ For the below instructions, replace `$MY_ACC_NUM` with your AWS account number a
...
-helm install rancher-csp-adapter rancher-charts/rancher-csp-adapter --namespace cattle-csp-adapter-system --set aws.enabled=true --set aws.roleName=$MY_ROLE_NAME --set-string aws.accountNumber=$MY_ACC_NUM
+helm install rancher-csp-adapter rancher-charts/rancher-csp-adapter --namespace cattle-csp-adapter-system --set aws.enabled=true --set aws.roleName=$MY_ROLE_NAME --set-string aws.accountNumber=$MY_ACC_NUM --set image.tag=v1.0.1
 ```
 
 
 Alternatively, you can use a `values.yaml` and specify options like below:
 
 ```yaml
+image:
+  tag: v1.0.1
```
cbron, line 76: This specific part looks weird, because it doesn't include repository. Would helm inherit that correctly if you are overriding the parent node ? Should be easy to test with `helm template` command.

MbolotSuse: Seems to work fine. I do need to update this section for another reason (`accountNumber` vs `accountNum`), so if you have an alternative I'm open to suggestions. That being said, if I specify repo here it makes this more complicated since I need to satisfy both the EU and non-eu listing.

cbron: As long as it inherits correctly then we should be fine.

cbron (approved):

MbolotSuse:
@jtravee Can you merge this when you get a chance?

MbolotSuse:
Thank you @btat

PR #4286, opened 2022-08-30
Adding docs/RN for all versions

jtravee:


src/pages/versions.md
```diff
@@ -0,0 +1,174 @@
...
+  <tr>
+    <td><b>v2.5.16</b></td>
+    <td><a href="https://rancher.com/docs/rancher/v2.5/en/">Documentation</a></td>
+    <td><a href="https://github.com/rancher/rancher/releases/tag/v2.5.16">Release Notes</a></td>
+  </tr>
+</table>
+
+<br/>
+
+### Past versions (no longer maintained)
```
btat, line 29:
```suggestion
### Past versions
```

src/pages/versions.md
```diff
@@ -0,0 +1,174 @@
+---
+title: Rancher Documentation Versions
+---
+
+### Current versions (stable)
```
btat, line 5:
```suggestion
### Current versions
```

src/pages/versions.md
```diff
@@ -0,0 +1,174 @@
...
+    <td><a href="https://github.com/rancher/rancher/releases/tag/v2.6.8">Release Notes</a></td>
+  </tr>
+</table>
+
+Below are the documentation and release notes for the currently released version of Rancher 2.5.x:
+
+<table>
+  <tr>
+    <td><b>v2.5.16</b></td>
+    <td><a href="https://rancher.com/docs/rancher/v2.5/en/">Documentation</a></td>
```
btat, line 22:
```suggestion
    <td><a href="https://URL_PLACEHOLDER/v2.5/rancher-manager">Documentation</a></td>
```

src/pages/versions.md
```diff
@@ -0,0 +1,174 @@
...
+<br/>
+
+### Legacy versions (EOL)
+
+Below is the documentation for legacy versions of Rancher 2.0 - 2.4.x:
+
+<table>
+  <tr>
+    <td><b>v2.0 - v2.4</b></td>
+    <td><a href="https://rancher.com/docs/rancher/v2.0-v2.4/en/">Documentation</a></td>
```
btat, line 172:
```suggestion
    <td><a href="https://URL_PLACEHOLDER/v2.5/rancher-manager">Documentation</a></td>
```

src/pages/versions.md
```diff
@@ -0,0 +1,174 @@
...
+---
+
+### Current versions (stable)
+
+Below are the documentation and release notes for the currently released version of Rancher 2.6.x:
+
+<table>
+  <tr>
+    <td><b>v2.6.8</b></td>
+    <td><a href="https://rancher.com/docs/rancher/v2.6/en/">Documentation</a></td>
```
btat, line 12:
```suggestion
    <td><a href="https://URL_PLACEHOLDER/">Documentation</a></td>
```

btat (changes requested):
The stable/unmaintained labels should be removed because of how our versions are structured. E.g. For 2.6.x the docs are either all maintained or all unmaintained, but it wouldn't be the case we maintain 2.6.8, but not earlier because our setup doesn't differentiate.

I've added examples of showing the how the known portions of the URLs will be formatted. Those will need to be updated for each entry.

Another column we could link to is the Support Matrix. Thoughts on adding that?

src/pages/versions.md
jtravee: Updated.

src/pages/versions.md
jtravee: Updated.

src/pages/versions.md
jtravee: Updated.

src/pages/versions.md
jtravee: Updated.

jtravee:
> The stable/unmaintained labels should be removed because of how our versions are structured. E.g. For 2.6.x, the docs are either all maintained or all unmaintained, but it wouldn't be the case we maintain 2.6.8, but not earlier because our setup doesn't differentiate. So it doesn't make sense to use the stable/maintained labels.
> 
> I've added examples of showing the how the known portions of the URLs will be formatted. Those will need to be updated for each entry.
> 
> Another column we could link to is the Support Matrix. Thoughts on adding that?

Great idea on the support matrix; I'll add those.

btat (changes requested):
Now that we have a confirmed URL, could you update the placeholders.

btat (changes requested):
The wrong URL was added. Should be `http://docs.ranchermanager.rancher.io/`.

Also, looks like parts of the URL got left off for the older versions.
E.g. /v2.5/ and /v2.0-v2.4/

btat (approved):

PR #4291, opened 2022-09-02
Reversed order of tabs per request

jtravee:
Updating order of tabs per request in 2.5, 2.6 per request in [Slack](https://suse.slack.com/archives/C02DQMPRA1E/p1662064439092729?thread_ts=1662063843.862589&cid=C02DQMPRA1E).

### DIFF
--- a/content/rancher/v2.5/en/admin-settings/authentication/azure-ad/_index.md
+++ b/content/rancher/v2.5/en/admin-settings/authentication/azure-ad/_index.md
@@ -4,24 +4,6 @@ weight: 1115
 ---
 
 {{% tabs %}}
-{{% tab "Rancher v2.5.0 - v2.5.15" %}}
-
-## Azure AD Graph API 
-
->**Important:** 
->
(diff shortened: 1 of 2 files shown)


maxsokolovsky (commented):
Thanks for the PR, @jtravee! I am having some difficulties with setting up the docs server locally in my environmental at the moment. I just wanted to check if anchored links work now.

For example, does visiting https://rancher.com/docs/rancher/v2.6/en/admin-settings/authentication/azure-ad/#3-set-required-permissions-for-rancher
take users to the right place? Does it work on your end?

jtravee:
> Thanks for the PR, @jtravee! I am having some difficulties with setting up the docs server locally in my environmental at the moment. I just wanted to check if anchored links work now.
> 
> For example, does visiting https://rancher.com/docs/rancher/v2.6/en/admin-settings/authentication/azure-ad/#3-set-required-permissions-for-rancher take users to the right place? Does it work on your end?

Yes, in my local test environment after the tab switch, this link works!

PR #4306, opened 2022-09-09
Updating Hugo banner for new docs site launch

jtravee:
To be merged after launch of new site.


btat:
This is not needed since we've added redirects and users wouldn't see this page anymore.

PR #4319, opened 2022-09-16
Fixes paths to product logos.

colstrom:
This should fix the logos.

### DIFF
--- a/content/_index.html
+++ b/content/_index.html
@@ -79,7 +79,7 @@ ctaBanner: true
                                                                                                         card-item">
               <div class="card-wrap">
                 <div class="image-container">
-                  <img class="image limited-height" alt="Harvester Logo" src="docs/harvester-logo.svg">
+                  <img class="image limited-height" alt="Harvester Logo" src="/docs/harvester-logo.svg">
                 </div>
                 <hr />
                 <p class="description-label">Harvester is an open-source hyper-converged infrastructure (HCI) software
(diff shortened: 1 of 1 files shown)


content/_index.html
```diff
@@ -79,7 +79,7 @@ <h2>Rancher Products</h2>
                                                                                                         card-item">
               <div class="card-wrap">
                 <div class="image-container">
-                  <img class="image limited-height" alt="Harvester Logo" src="docs/harvester-logo.svg">
+                  <img class="image limited-height" alt="Harvester Logo" src="/docs/harvester-logo.svg">
```
mattfarina, line 82: When published live, this path would translate to https://rancher.com/docs/harvester-logo.svg. The logo isn't there. It's at https://rancher.com/docs/docs/harvester-logo.svg

colstrom: Fabulous catch, sir. I'll fix that!

content/_index.html
```diff
@@ -79,7 +79,7 @@ <h2>Rancher Products</h2>
                                                                                                         card-item">
               <div class="card-wrap">
                 <div class="image-container">
-                  <img class="image limited-height" alt="Harvester Logo" src="docs/harvester-logo.svg">
+                  <img class="image limited-height" alt="Harvester Logo" src="/docs/docs/harvester-logo.svg">
```
mattfarina, line 82: This breaks in dev environments where the images are at /docs/[img]. The docs dev environment is nested one level deep compared to rancher.com.

catherineluse: Oh, I see what happened. This is my fault. The links on this page used to use the `baseUrl` variable so that dev and prod would use different directories, and I didn't realize that changed it without thinking. These links should be formatted like this: `src="{{< baseurl >}}/docs/harvester-logo.svg/">`

catherineluse (changes requested):

mattfarina:
I created https://github.com/rancher/docs/pull/4321 as an alternative fix to consider.

content/_index.html
colstrom: Ah, excellent. I wasn't aware of `{{< baseurl >}}`. I'll amend accordingly. Is that something provided by Hugo?

btat:
It looks like the line with "rancher-logo" was missed.

content/_index.html
catherineluse: yes

PR #4320, opened 2022-09-16
Fix relative path

mattfarina:
This is an alternative solution to #4319 which uses nginx to redirect /docs to /docs/. This should enable relative paths to have the correct root base in their path.


mattfarina:
dang it... closign to re-open

PR #4343, opened 2022-10-25
add action to release docs

kshtsk:


btat:
Hi @kshtsk, Guna mentioned that the use of GHCR is for testing purposes only. Given that's the case, is it correct to assume removal of the existing drone.yml happens at a later time?

kshtsk:
> Hi @kshtsk, Guna mentioned that the use of GHCR is for testing purposes only. Given that's the case, is it correct to assume removal of the existing drone.yml happens at a later time?

Hi @btat 
I don't know about what are the plans, what I understand it is going to move this part of infra to a new cluster, and one of the option of improving the process is to migrate building of containers closer to the respecting repos, in particular, use the actions as much as possible, which _could_ follow to discontinue drone usage, so as soon as 'actions' work the `.drone.yml` can be dropped. Since building containers is going to happen in actions, then it is reasonable to push images to GHCR. Later on we can switch infra to use GHCR instead dockerhub or other, however it is not decided it yet, as I understood from Paul's words. Now, since actions do not interfere the drone jobs, there is no acute need to drop its config now.

PR #4377, opened 2023-10-13
Update Cilium CNI Popularity Chart Numbers

amitmavgupta:
Update Cilium CNI Popularity Chart Numbers


gunamata:
@btat , I think this needs to be updated in the current (Docusaurus)docs? What's the corresponding page by the way in current docs?

btat:
> @btat , I think this needs to be updated in the current (Docusaurus)docs? What's the corresponding page by the way in current docs?

I will port this over to the current (Docusaurus) docs and confirm the corresponding page as part of it. But we'll need to re-evaluate this table and modify it because the data we list constantly changes.

btat:
Replaced by https://github.com/rancher/rancher-docs/pull/992 on the current docs repo.

